PR #1, opened 2015-10-20
[wip] remove required kw args to support ruby 2.0

simon3z:
This is my temporary workaround to work with ruby 2.0. It probably requires discussion on whether we want to support ruby 2.0 and if so, what may be missing in this patch.


lzoubek:
I don't have strong opinion of versions ruby to support (older versions seem to be better). If you need to run client on 2.0 then it makes sense to support it.

pilhuhn:
@lzoubek I think the issue is rather the other way around. In 2.0 there is no support for required keywords (but the old hack hash still exists). I am +1 for supporting 2.0

PR #3, opened 2015-10-20
Http add headers

simon3z:
This is based on #2. Let's tackle that one first.


lzoubek:
if I merged #2 I won't be able to merge #3 (I think) as #2 is subset of #3. Merging, thank you

PR #5, opened 2015-10-23
add support for bucketDuration in get_data

simon3z:
cc @lzoubek @pilhuhn


simon3z:
ping @lzoubek @pilhuhn

PR #11, opened 2015-10-28
client: support token authorization

simon3z:
cc @pilhuhn


pilhuhn:
Can you please also update the spec, as this gives "compile" errors

pilhuhn:
All Travis failures are due to non-existing Hawkular live instance

PR #12, opened 2015-10-28
lib: support rubocop coding style checks

simon3z:
cc @pilhuhn

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,12 @@
+MethodLength:
+  Enabled: false
+ClassLength:
+  Enabled: false
+ModuleLength:
+  Enabled: false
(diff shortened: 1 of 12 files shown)


pilhuhn:
This prevents the rubocop errors locally:

```
def tenant_header
  @options[:tenant].nil? ? {} : {
    :'Hawkular-Tenant' => @options[:tenant]
  }
end
```

Note that `tenantId` header is old and not used. Only Hawkular-Tenant - see http://www.hawkular.org/docs/rest/rest-metrics.html#_tenant_header

pilhuhn:
I think this is now resolved by my earlier pr, @simon3z ?

simon3z:
Correct @pilhuhn thanks!

PR #15, opened 2015-11-05
metrics: allow query without tags to list all resources

simon3z:
cc @pilhuhn

### DIFF
--- a/lib/metrics/metric_api.rb
+++ b/lib/metrics/metric_api.rb
@@ -60,8 +60,9 @@ module Hawkular::Metrics
       # Query metric definitions by tags
       # @param tags [Hash]
       # @return [Array[MetricDefinition]]
-      def query(tags)
-        @client.http_get("/metrics/?type=#{@type}&tags=#{tags_param(tags)}").map do |g|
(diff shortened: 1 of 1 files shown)


lib/metrics/metric_api.rb
```diff
@@ -60,8 +60,9 @@ def create(definition)
       # Query metric definitions by tags
       # @param tags [Hash]
       # @return [Array[MetricDefinition]]
-      def query(tags)
-        @client.http_get("/metrics/?type=#{@type}&tags=#{tags_param(tags)}").map do |g|
+      def query(tags = nil)
+        tags_filter = tags.nil? ? '' : "&tags=#{tags_param(tags)}"
+        @client.http_get("/metrics/?type=#{@type}#{tags_filter}").map do |g|
```
pilhuhn, line 65: Should that really append the # in case that there is no filter specified?
What do you think @burmanm ?

pilhuhn, line 65: Ok, my bad did not recognize `#{}` as such, as I was too much focused of the # in html-links to anchors.

PR #20, opened 2016-02-16
Added token support.

jpkrohling:


jpkrohling:
This is compatible with Hawkular Alpha 10, but not Alpha 9 (there was a change in the REST endpoints for the secret store). Once Alpha 10 is released, I'll remove the "Don't merge yet" label.

pilhuhn:
Can you please also add a "negative test" for invalid user/invalid persona?
What is the relation here with standalone Hawkular-Metrics - any impact (I doubt it, asking anyway)?

jpkrohling:
Updated! There's no relation with standalone Hawkular-Metrics, as there's no auth there.

PR #21, opened 2016-02-18
Extending the inventory client

jkremser:
- fixing some itests
- adding a class for working with canonical path
- extending the inventory client

~~todo: record/create vcr for `list_metrics_for_metric_type` and `list_metrics_for_resource_type` methods
=> don't merge yet~~
done


lib/inventory/inventory_api.rb
```diff
@@ -56,15 +54,13 @@ def list_resource_types(feed = nil)
...
-      ret.each { |rt| val.push(ResourceType.new(rt)) }
-      val
+      ret.map { |rt| ResourceType.new(rt) }
     end
 
     # List the resources for the passed feed and resource type. The representation for
     # resources under a feed are sparse and additional data must be retrived separately.
     # It is possible though to also obtain runtime properties by setting #fetch_properties to true.
-    # @param [String] feed The id of the feed the type lives under. Can be nil for feedless types
+    # @param [String] feed The id of the feed the type lives under. Can be nil for all feeds
```
pilhuhn, line 63: I was using the word feedless to indicate that things like the URLs in Hawkular do not have a feed as opposed to e.g. WildFlys to have a feed. My wording was probably not good - we need something better.

lib/inventory/inventory_api.rb
```diff
@@ -166,9 +160,50 @@ def list_relationships_for_feed(feed_id)
...
+
+    # List the metrics for the passed feed and all the resources of given resource type.
+    # If feed is not passed, all the metrics across all the feeds of a resource type will be retrieved
+    # This method may perform multiple REST calls.
+    # @param [String] feed The id of the feed the type lives under. Can be nil for all feeds
+    # @param [String] resource_type Name of the resource type to look for. Can be obtained from {ResourceType}.id.
+    #   Must not be nil
+    # @return [Array<Metric>] List of metrics. Can be empty
+    def list_metrics_for_resource_type(feed, _resource_type)
+      fail 'Type must not be nil' unless type
```
pilhuhn, line 201: There is no parameter `type`

spec/integration/inventory_spec.rb
```diff
@@ -11,8 +11,7 @@ module Hawkular::Inventory::RSpec
 
       tenant = client.get_tenant(creds)
 
-      expect tenant.nil?
-      expect tenant.eql?('28026b36-8fe4-4332-84c8-524e173a68bf')
+      expect(tenant).to eq('28026b36-8fe4-4332-84c8-524e173a68bf')
```
pilhuhn, line 17: Good catch!

lib/inventory/inventory_api.rb
jkremser, line 63: Right, I have the same understanding. Feedless ~ without feed (directly under environment) vs under feed. The reason, I've changed this is that

``` bash
curl -u jdoe:password http://localhost:8080/hawkular/inventory/resourceTypes
```

actually fetches bot resource types that are reported under the feed (WF, OS, ..) and without the feed (just URL)

There is also a query param for fethcing only the feedless ones `?feedless=true`

lib/inventory/inventory_api.rb
jkremser, line 201: good catch

lib/inventory/inventory_api.rb
```diff
@@ -375,4 +409,86 @@ def to_h
...
+      ret += "/rt;#{@resource_type_id}" unless @resource_type_id.nil?
+      ret += "/mt;#{@metric_type_id}" unless @metric_type_id.nil?
+      ret += "/m;#{@metric_id}" unless @metric_id.nil?
+      ret += resources_chunk.to_s
+      ret
+    end
+
+    private
+
+    def self.parse_other(key, val, hash)
```
jkremser: btw. this is here because of the Rubocop complaining about cyclomatic complexity > 6 . So the `self.parse` handles the first 3 cases and this method the rest 4.. I know, this smells. But I didn't find any other way to do that. Perhaps there could be an exception for the `case-when` statements, because it would me more readable to have it all in 1 method.

pilhuhn: On 18 Feb 2016, at 11:31, Jirka Kremser wrote:

> > -    end
> >   +
> > -    def to_s
> > -      ret = "/t;#{@tenant_id}"
> > -      ret += "/f;#{@feed_id}" unless @feed_id.nil?
> > -      ret += "/e;#{@environment_id}" unless @environment_id.nil?
> > -      ret += "/rt;#{@resource_type_id}" unless 
> >   @resource_type_id.nil?
> > -      ret += "/mt;#{@metric_type_id}" unless @metric_type_id.nil?
> > -      ret += "/m;#{@metric_id}" unless @metric_id.nil?
> > -      ret += resources_chunk.to_s
> > -      ret
> > -    end
> >   +
> > -    private
> >   +
> > -    def self.parse_other(key, val, hash)
> 
> btw. this is here because of the Rubocop complaining about cyclomatic 
> complexity > 6 . So the `self.parse` handles the first 3 cases and 
> this method the rest 4.. I know, this smells. But I didn't find any 
> other way to do that. Perhaps there could be an exception for the 
> `case-when` statements, because it would me more readable to have it 
> all in 1 method.

Yes, don't play such tricks.
there is a `.rubocop` file that can be tuned.

lib/inventory/inventory_api.rb
```diff
@@ -105,47 +100,49 @@ def get_config_data_for_resource(resource_id, feed)
...
       the_feed = hawk_escape parent_resource.feed
       the_id = hawk_escape parent_resource.id
 
       ret = http_get('/feeds/' + the_feed +
-                     '/resources/' + the_id + '/children')
-      val = []
-      ret.each { |r| val.push(Resource.new(r)) }
-      val
+                     '/resources/' + the_id +
+                     (recursive ? '/recursiveChildren' : '/children'))
```
pilhuhn, line 112: Externalize to `generate_query_params` as you did above?

lib/inventory/inventory_api.rb
```diff
@@ -105,47 +100,49 @@ def get_config_data_for_resource(resource_id, feed)
...
+      ret.map { |r| Relationship.new(r) }
     rescue
       []
     end
 
+    # Retrieve a single entity from inventory by its canonical path
+    # @param [String] path canonical path of the entity
+    # @return inventory entity
+    def get_entity(path)
+      http_get("path#{path}")
```
pilhuhn, line 143: Is this missing the initial `/` in front of the path, or is the path guaranteed to start with a slash?
I also did not see a spec for it (but may have just missed it).

lib/inventory/inventory_api.rb
jkremser, line 112: this is not query, it's the elvis operator

pilhuhn, line 112: > this is not query, it's the elvis operator

Right, sorry for the noise.

lib/inventory/inventory_api.rb
jkremser, line 143: this works, I'll add to the docs that it only accept the absolute path (i.e. starting with /)
or it can actually accept the CanonicalPath instance to provide some kind of "type safety"

PR #24, opened 2016-02-22
More inventory work

pilhuhn:


pilhuhn:
@jpkrohling @Jiri-Kremser Can you have a look please?

lib/inventory/inventory_api.rb
```diff
@@ -57,6 +57,23 @@ def list_resource_types(feed = nil)
       ret.map { |rt| ResourceType.new(rt) }
     end
 
+    # Return all resources for a feed
+    # @param [String] feed Id of the feed that hosts the resources
+    # @param [Boolean] fetch_properties Should the config data be fetched too
+    # @return [Array<Resource>] List of resources, which can be empty.
+    def list_resources_for_feed(feed, fetch_properties = false)
+      fail 'Feed must be given' unless feed
+      the_feed = hawk_escape(feed)
```
jpkrohling: In some places, there are parenthesis when calling hawk_escape, in other places it's without.

Also, wouldn't it be a good candidate for a monkey patching, specially if this escape is used in other non-inventory places? No need to make it system wide, though.

lib/inventory/inventory_api.rb
```diff
@@ -246,8 +263,186 @@ def list_metrics_for_resource(resource, filter = {})
...
+    # @param [String] feed_id  Id of a feed - required
+    # @param [String] feed_name A display name for the feed
+    # @return [Object]
+    def create_feed(feed_id, feed_name = nil)
+      feed = create_blueprint
+      feed[:id] = feed_id
+      feed[:name] = feed_name
+
+      begin
+        return http_post('/feeds/', feed)
```
jpkrohling, line 273: It seems that this return isn't needed, is it?

lib/hawkular.rb
```diff
@@ -50,6 +50,8 @@ def http_get(suburl, headers = {})
     # @param [String] url_part Part of an url to be escaped
     # @return [String] escaped url_part as new string
     def hawk_escape(url_part)
+      return "#{url_part}" if url_part.is_a?(Numeric)
```
jpkrohling: Why are you wrapping into a string with vars? You could just use `url_part.to_s`, no?

lib/inventory/inventory_api.rb
```diff
@@ -246,8 +263,186 @@ def list_metrics_for_resource(resource, filter = {})
...
+    # @return [ResourceType] ResourceType object just created
+    def create_resource_type(feed_id, type_id, type_name)
+      the_feed = hawk_escape feed_id
+
+      type = create_blueprint
+      type[:id] = type_id
+      type[:name] = type_name
+
+      begin
+        res = http_post('/feeds/' + the_feed + '/resourceTypes', type)
```
jpkrohling: I guess it's a matter of taste, but why not use double quotes + vars instead of three concatenated strings? Ie:

`http_post("/feeds/#{the_feed}/resourceTypes", type)`

lib/inventory/inventory_api.rb
```diff
@@ -246,8 +263,186 @@ def list_metrics_for_resource(resource, filter = {})
...
+      rescue => error
+        # 409 We already exist -> that is ok
+        if error.status_code == 409
```
jpkrohling, line 308: Will `error` always have a `status_code` at this point? Wouldn't it be better to do a `if error.responds_to?(:status_code) and error.status_code == 409` ? Unless you change the `rescue` to rescue a specific error.

lib/inventory/inventory_api.rb
```diff
@@ -246,8 +263,186 @@ def list_metrics_for_resource(resource, filter = {})
...
+      end
+
+      res = http_get("/feeds/#{the_feed}/metricTypes/#{metric_type_id}")
+
+      MetricType.new(res)
+    end
+
+    def build_metric_type_hash(collection_interval, metric_kind, metric_type_id, unit)
+      mt = {}
+      mt['id'] = metric_type_id
```
jpkrohling, line 390: Is there a reason for using string as the hash keys?

lib/inventory/inventory_api.rb
```diff
@@ -246,8 +263,186 @@ def list_metrics_for_resource(resource, filter = {})
...
+    # @param [String] metric_id Id of the metric
+    # @param [String] type_path Full path of the MetricType
+    # @param [String] resource_id Id of the resource to associate the metric with
+    # @return [Metric] The metric created or if it already existed the version from the server
+    def create_metric_for_resource(feed_id, metric_id, type_path, resource_id)
+      the_feed = hawk_escape feed_id
+      the_resource = hawk_escape resource_id
+
+      m = {}
+      m['id'] = metric_id
```
jpkrohling, line 408: Same as above (hash keys as string)

lib/inventory/inventory_api.rb
```diff
@@ -246,8 +263,186 @@ def list_metrics_for_resource(resource, filter = {})
...
+      res[:properties] = properties
+      res[:id] = resource_id
+      res[:name] = resource_name
+      res[:resourceTypePath] = type_path
+
+      begin
+        http_post('/feeds/' + the_feed + '/resources', res)
+      rescue => error
+        # 409 We already exist -> that is ok
+        raise unless error.status_code == 409
```
jpkrohling, line 339: Same as above: restrict the error type, or ensure the `status_code` exists.

lib/inventory/inventory_api.rb
```diff
@@ -246,8 +263,186 @@ def list_metrics_for_resource(resource, filter = {})
...
+      metric_kind = type.nil? ? 'GAUGE' : type.upcase
+      fail "Unknown type #{metric_kind}" unless %w(GAUGE COUNTER AVAILABILITY').include?(metric_kind)
+
+      mt = build_metric_type_hash(collection_interval, metric_kind, metric_type_id, unit)
+
+      begin
+        http_post("/feeds/#{the_feed}/metricTypes", mt)
+      rescue => error
+        # 409 We already exist -> that is ok
+        raise unless error.status_code == 409
```
jpkrohling, line 380: Same as above: restrict the error type, or ensure the status_code exists.

lib/inventory/inventory_api.rb
```diff
@@ -246,8 +263,186 @@ def list_metrics_for_resource(resource, filter = {})
...
+      end
+
+      ret = http_get("/feeds/#{the_feed}/metrics/#{metric_id}")
+      the_metric = Metric.new(ret)
+
+      begin
+        http_post("/feeds/#{the_feed}/resources/#{the_resource}/metrics", [the_metric.path])
+      rescue => error
+        # 409 We already exist -> that is ok
+        raise unless error.status_code == 409
```
jpkrohling, line 425: Same as previously: restrict the error type, or ensure the status_code exists.

lib/inventory/inventory_api.rb
```diff
@@ -348,6 +545,41 @@ def to_h
...
+    # @return [String] Name of the metric
+    attr_reader :name
+    attr_reader :id
+    attr_reader :feed
+    attr_reader :unit
+    # @return [Long] collection interval in seconds
+    attr_reader :collection_interval
+
+    def initialize(type_hash)
+      @id = type_hash['id']
```
jpkrohling, line 557: Same as a previous comment: hash keys as string

lib/inventory/inventory_api.rb
```diff
@@ -348,6 +545,41 @@ def to_h
...
+      @path = type_hash['path']
+      @name = type_hash['name'] || @id
+      @_hash = type_hash.dup
+
+      tmp = path.split('/')
+      tmp.each do |pair|
```
jpkrohling, line 563: Doesn't/shouldn't rubocop complain about a lack of space around `pair` on this case?

lib/inventory/inventory_api.rb
```diff
@@ -348,6 +545,41 @@ def to_h
...
+        (key, val) = pair.split(';')
```
jpkrohling, line 564: I might be wrong here, but I think `String#split`  receives a `Regexp` by default. So, you could use `/;/` instead of letting it accept a string and transform into a Regexp.

lib/inventory/inventory_api.rb
```diff
@@ -246,8 +263,186 @@ def list_metrics_for_resource(resource, filter = {})
...
+          return http_get('/feeds/' + the_feed)
+        else
+          raise
+        end
+      end
+    end
+
+    # Delete the feed with the passed feed id.
+    # @param feed Id of the feed to be deleted.
+    def delete_feed!(feed)
```
jpkrohling: Why the exclamation mark here? If there's an exclamation mark here, shouldn't there also be one for the `create_feed` method?

jpkrohling:
Review finished. I got a couple of questions, but it looks good to me. However, I cannot judge the inventory-specific things.

lib/inventory/inventory_api.rb
pilhuhn: I put the exclamation mark here, to indicate that this is a dangerous method.

lib/inventory/inventory_api.rb
pilhuhn, line 557: Yes, because it fails with a symbol. I think this is because he input is a hash that was obtained from the remote and then translated to a hash via `JSON.parse(res)`

lib/inventory/inventory_api.rb
pilhuhn, line 308: Good point

lib/inventory/inventory_api.rb
jpkrohling: IIRC, "dangerous method" is defined in Ruby as methods that change the instance itself, which is not the case here.

lib/inventory/inventory_api.rb
pilhuhn, line 564: It receives a "something" and then reacts on its type. The results are different for splitting on `' '` vs `/ /` but that does not matter here. So I think it is equivalent in this case.

PR #28, opened 2016-03-03
- Add list-events support in alerts-api

jshaughn:
- Add to the spec
  - note, see mechanism for ignoring parts of the uri in the vcr match
- refactor generate_query_params
  - not really necessary, mainly a ruby exercise for me

### DIFF
--- a/lib/alerts/alerts_api.rb
+++ b/lib/alerts/alerts_api.rb
@@ -2,8 +2,10 @@ require 'hawkular'
 require 'ostruct'
 
 # Alerts module provides access to Hawkular-Alerts.
-# There are two main parts here: Triggers, that define alertable conditions
-# and Alerts that represent a fired trigger.
+# There are three main parts here:
+#   Triggers, that define alertable conditions
+#     Alerts, that represent a fired Alert trigger
+#     Events, that represent a fired Event trigger or an externally injected event (hawkular, not miq event)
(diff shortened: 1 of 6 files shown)


lib/alerts/alerts_api.rb
```diff
@@ -67,9 +69,11 @@ def get_alerts_for_trigger(trigger_id) # TODO: add additional filters
     end
 
     # List fired alerts
+    # @param [Hash]criteria optional query criteria
```
pilhuhn, line 72: Can you add the key, value types for the hash please?

lib/alerts/alerts_api.rb
```diff
@@ -109,6 +113,17 @@ def acknowledge_alert(alert_id, by = nil, comment = nil)
...
       true
     end
+
+    # List Events given optional criteria
+    # @param [Hash] criteria optional query criteria
+    # @return [Array<Event>] List of all alerts in the system. Can be empty
+    def list_events(*criteria)
+      query = generate_query_params(*criteria)
+      ret = http_get('/events' + query)
+      val = []
```
pilhuhn: Try to use `ret.map{..}`. I know I've used this current way also (too often).

pilhuhn:
Two small comments - otherwise looks good to me :+1:

jshaughn:
I amended the prior commit and force-pushed, so comment lines may be off, but the changes are in.

PR #30, opened 2016-03-08
Operations

jkremser:
- [x]  generic operations like Undeploy, Redeploy, Enable, Disable, Remove Datasource, Remove Driver, etc.
- [x] tests 
- [x] operations with binary data like AddDeployment, AddDatasource
- [x] AddJDBCDriver
- [ ] ~~export JDR~~ 

Unfortunatelly, the `Expord JDR` is not doable with the `websocket-client-simple`. It's reading the incoming messages char by char from the socket, but the JDR report is pretty big. I've created a PR on the client, but this doesn't help much. We need to find out something else. Perhaps better than sending the bits over websocket, we could have send only a link where to download it.


pilhuhn:
"export jdr" is an example of an operation that is fetching binary data from agent->hawkular->Client. I think we will need that going forward.

theute:
JDR support is in the "PRD"

spec/integration/inventory_spec.rb
```diff
@@ -105,8 +100,7 @@ module Hawkular::Inventory::RSpec
     it 'Should List datasources with no props' do
       creds = { username: 'jdoe', password: 'password' }
 
-      client = Hawkular::Inventory::InventoryClient.new(INVENTORY_BASE, creds)
```
pilhuhn: Please back those changes out and submit as a separate PR.

spec/vcr/vcr_setup.rb
```diff
@@ -8,9 +8,3 @@
   c.debug_logger = File.open('spec/vcr.log', 'w')
   c.configure_rspec_metadata!
 end
-
-RSpec.configure do |c|
-  # so we can use :vcr rather than :vcr => true;
-  # in RSpec 3 this will no longer be necessary.
```
pilhuhn, line 14: Why did you remove this?

jkremser, line 14: I was getting deprecation warning every time I run rspec. This is related https://git.io/vavzJ

spec/integration/inventory_spec.rb
jkremser: done, pr is here
https://github.com/hawkular/hawkular-client-ruby/pull/31

lib/operations/operations_api.rb
```diff
@@ -0,0 +1,144 @@
+require 'hawkular'
+require 'websocket-client-simple'
+require 'json'
+
+class Proc
```
pilhuhn, line 6: Are you sure that this should go into the root namespace and not under Hawkular::Operations ?

Also: documentation is missing

jkremser, line 6: I am defining a perform `method()` on any block (top lvl) that can have multiple methods accepting another nested blocks. Currently the nested blocks are `.success` and `.failure` which are the actuall callbacks) Code is quite complex, but the resulting API is imho easy to read:

``` ruby
client = OperationsClient.new(...)
client.invoke_generic_operation(foo) do |on|
  on.success do |data|
    # do something with data
  end
  on.failure do |error|
    # handle error
  end
end
```

I've added some docs and check that the blocks respond_to? 'perform', actually, the PR was pretty broken by my manual split and git auto merge. Tests were not passing.

I haven't found a way to make it working under the `Hawkular::Operations` module. If the class definition/overriding is inside the `Hawkular::Operations module`, it would be talking about `Hawkular::Operations::Proc` and not the `Proc`

PR #31, opened 2016-03-08
simplifying the inventory spec

jkremser:


lib/inventory/inventory_api.rb
```diff
@@ -9,13 +9,13 @@ module Hawkular::Inventory
...
-    #   http://localhost:8080/hawkular/inventory
-    # @param credentials [Hash{String=>String}] Hash of username, password, token(optional)
-    def initialize(entrypoint = nil, credentials = {})
-      @entrypoint = entrypoint
-
-      super(entrypoint, credentials)
+    # @param hash [Hash{String=>Object}] a hash containing base url of Hawkular-inventory - e.g
+    #   entrypoint: http://localhost:8080/hawkular/inventory
+    # and another sub-hash containing the hash with username, password, token(optional)
+    def initialize(hash)
```
pilhuhn, line 15: This was actually a bad change for 0.2.1 as this method is already in use by clients.

jkremser, line 15: I can create a PR to MiQ fixing this

pilhuhn, line 15: Who tells you that MiQ is the only client? Within 0.2.\* we must not introduce such a change.

jkremser, line 15: Right, the next release should be IMHO 2.3.0, because we are adding features (not just bug fixing). But still, this is a breaking change. Ok, so I'll make it compatible again. Unfortunately, we can't have more constructors in Ruby so I can't add the previous one and have both, in general I like hashes as parameters more than normal parameters with default values, because it's more obvious during the instance creation what parameter belongs to what, that's why I did this change. There was actually no way to provide only credentials and leave the entrypoint fall back to default, because the parameters are ordered you can't provide only 2nd argument without providing the 1st one.

btw. this is the change needed on MiQ side (https://gist.github.com/Jiri-Kremser/df55e0c450122d9b6304)

PR #33, opened 2016-03-14
More work on alert trigger handling.

pilhuhn:
That is certainly not complete (missing documentation).

@jshaughn and @lponce would you have a look please?

### DIFF
--- a/lib/alerts/alerts_api.rb
+++ b/lib/alerts/alerts_api.rb
@@ -55,10 +55,19 @@ module Hawkular::Alerts
       trigger
     end
 
-    def bulk_load_triggers(hash)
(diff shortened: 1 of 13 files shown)


lib/alerts/alerts_api.rb
```diff
@@ -55,6 +55,63 @@ def get_single_trigger(trigger_id, full = false)
       trigger
     end
 
+    def bulk_load_triggers(hash)
```
jkremser: I am working on this PR to make it mergeable and I'd rename this method to `bulk_import_triggers` because `load` evokes the server -> client communication for me, but /me not speaking good English so @jshaughn wdyt? :]

Issue #41, opened 2016-03-23
Add offline tests for websockets


pilhuhn:
Websockets are currently not offline-tested as VCR can't do that.

pilhuhn:
Done via #44

PR #44, opened 2016-03-24
issue #41 - Add offline tests for websockets

jkremser:
This adds a development dependency on `simple-websocket-vcr`. 

It stores the cassettes with the websocket communication in a json format as a 2d array The outer array stores various websocket sessions, because there can be multiple connections per 1 cassete and the inner array is for particular operations.

currently, available operations are: 
- `read` (server -> client msg), 
- `write`
- and `close`

Each operation has also the data type, because we need to recognize the binary websocket frames and the textual ones. Binary data is encoded by `Base64`

Hopefully, this should return the coverage back to ~90


pilhuhn:
This fails on Ruby 2.0 - can you please check @Jiri-Kremser 
Otherwise great Job! :+1:

PR #49, opened 2016-04-11
Support for ERB templating in inventory specs

jkremser:
`VCR_UPDATE=1` will now re-record the **templates** against the currently running Hawkular instance. By template I mean the recorded communication in yaml with `<%= foobar =>` in it. This works for both `VCR` (done manually here https://git.io/vV5PM) and `WebSocketVCR` (it's a feature that is turned on by `reverse_substitution: true`)

`VCR_OFF=1` is still working and can be used for running the tests against the live Hawkular  instance, w/o recording it

Point of the whole PR is not to over write the VCR tapes over and over again, because the feed_id is randomly generated, or for those UUIDs in tests.


spec/integration/inventory_spec.rb
```diff
@@ -25,24 +25,46 @@ module Hawkular::Inventory::RSpec
...
-      @state = {
-        hostname: 'localhost.localdomain',
-        feed: nil
-      }
+      options = { decode_compressed_response: true }
+      options[:record] = :all if ENV['VCR_UPDATE'] == '1'
+      VCR.use_cassette('Inventory/Helpers/get_feeds', options) do
+        feeds = @client.list_feeds
+        @state = {
+          hostname: 'dhcp-10-40-1-43.brq.redhat.com',
```
pilhuhn: Do I get this right, that the client still talks to Hawkular on e.g. localhost, and that this 'state' is a placeholder?

pilhuhn:
Not sure I get it. Where does `reverse_substitution: true` come from?
And how does this handle the `<% =bla %>` erb entries in an existing cassette? Be overwriting the existing cassette this erb info is lost?
WebsocketVCR is the `simple-websocket-vcr` that you have contributed to?

spec/integration/inventory_spec.rb
jkremser: > Do I get this right, that the client still talks to Hawkular on e.g. localhost

nope, it talks to hawkular only if `VCR_OFF=1` (use case: trying if it still works) or `VCR_UPDATE=1` (use case: updating the cassettes/templates, perhaps the hawk api was changed)

> ..and that this 'state' is a placeholder?

Yeah, it's kind of a placeholder so that the `feed_id` don't have to be hardcoded to a value that is stored in the cassette. Actually the `Inventory/Helpers/get_feeds` cassette is the only cassette with the `feed_id` all the others have it parametrized by the ERB.

wrt `hostname='dhcp-10-40-1-43.brq.redhat.com'`, it's probably wrong how it is right now, perhaps I should have removed this check, because it's different on each environment and this makes updating the cassetes or running the tests online a pain.

jkremser:
> Not sure I get it. Where does reverse_substitution: true come from?

Yes, this is an option for `WebsocketVCR` (simple-websocket-vcr gem)

> And how does this handle the <% =bla %> erb entries in an existing cassette? Be overwriting the existing cassette this erb info is lost?
> it doesn't, all it does is this:
- if the cassette with the given name is there (and `record` option **is not** set to `all`), it uses the cassette and no reverse substitution is done.
- if the cassette with the given name is there (and `record` option **is** set to `all`, it records the normal cassette with all the ids and other real data captured during the comm, but right before the saving it does the substitution so that all the `values` (from the `erb` hash) are replaced by the `<% =key %>`, so yes if this is the default setting it would rewrite the "template" cassette all the time, but with the same data, because all the differences will be handled by the `<%= varying_stuff %>`. What we do is, using this settings only if `VCR_UPDATE=1` (https://git.io/vwmE9)
- if the cassette with the given name doesn't exist, it does the same as previous case

> WebsocketVCR is the simple-websocket-vcr that you have contributed to?

yes

There is no such an option for normal VCR, so I cheated and did the similar thing as WebsocketVCR in the around "interceptor": 
- https://git.io/vwmz8
- https://git.io/vwmz2

jkremser:
oh here is basically what the reverse substituion is:

https://github.com/hawkular/hawkular-client-ruby/pull/49/files#diff-93830fa29d616f7c87903d08b5b1b29aR118

``` ruby
    bindings.select { |_, v| v.size >= 3 }.each do |k, v|
       text.gsub! v, "<%= #{k} %>"
    end
```

spec/integration/inventory_spec.rb
pilhuhn: When I said 'talks to' - I meant in case of (re-)recording. Because you changed 'localhost' to 'dhcp-...'
But seems the last paragraph you wrote goes into the direction I was thinking as well.

jkremser: yeah, the `hostname` isn't the ip/host of the hawkular into which the vcr is connecting, but the value that is expected in the JSON. I think it's stored in the configuration of wildfly resource. 

The tests currently run always against localhost (https://github.com/Jiri-Kremser/hawkular-client-ruby/blob/0.3.0/spec/integration/inventory_spec.rb#L10)

jkremser:
closing the PR, I'll re-open it against the master, since the `0.2.x` stream has it's own branch now

PR #56, opened 2016-04-21
Erb templating

jkremser:
This is basically the same PR as the #49 but now targeting directly the master branch (not the `0.3.0` integration branch)

`VCR_UPDATE=1` will now re-record the **templates** against the currently running Hawkular instance. By template I mean the recorded communication in yaml with `<%= foobar =>` in it. This works for both `VCR` (done manually here https://git.io/vV5PM) and `WebSocketVCR` (it's a feature that is turned on by `reverse_substitution: true`)

`VCR_OFF=1` is still working and can be used for running the tests against the live Hawkular  instance, w/o recording it

Point of the whole PR is not to over write the VCR tapes over and over again, because the feed_id is randomly generated, or for those UUIDs in tests.


pilhuhn:
LGTM - @abonas can you please have a look as well?

abonas:
I will review this on Sunday

CHANGES.rdoc
```diff
@@ -3,6 +3,10 @@
 This document describes the relevant changes between releases of the
 _hawkular-client_ project.
 
+=== V 0.3.0 (not released yet)
```
abonas: this section looks irrelevant for this PR?

README.rdoc
```diff
@@ -31,6 +31,10 @@ Integration tests are recorded and played against cassettes recorded with VCR
     rake spec
 * To run the tests against a live server, set +VCR_OFF+ to +1+ as in
     VCR_OFF=1 rake spec
+* To update the VCR tapes that supports it, set +VCR_UPDATE+ to +1+ as in
```
abonas: this section looks irrelevant for this PR?

hawkularclient.gemspec
```diff
@@ -27,7 +27,7 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency('shoulda')
   gem.add_development_dependency('rspec-rails', '~> 3.0')
   gem.add_development_dependency('rake', '< 11')
-  gem.add_development_dependency('simple-websocket-vcr', '= 0.0.4')
+  gem.add_development_dependency('simple-websocket-vcr', '= 0.0.7')
```
abonas, line 30: why does this PR require this gem's upgrade?

spec/integration/inventory_spec.rb
```diff
@@ -25,24 +25,45 @@ module Hawkular::Inventory::RSpec
...
+        }
+      end
+    end
+
+    after(:all) do
+      require 'fileutils'
+      FileUtils.rm_rf "#{VCR.configuration.cassette_library_dir}/Inventory/tmp"
+    end
+
+    let(:cassette_name) do |e|
```
abonas: what does `e` stand for?

spec/integration/inventory_spec.rb
```diff
@@ -25,24 +25,45 @@ module Hawkular::Inventory::RSpec
...
+      description = e.description
+      description
```
abonas, line 52: why is this line needed, and not just return e.description? (last line is the return value)

spec/integration/inventory_spec.rb
```diff
@@ -110,36 +131,36 @@ module Hawkular::Inventory::RSpec
...
 
     it 'Should list recursive children of WildFly' do
-      resources = @client.list_resources_for_type(@state[:feed], 'WildFly Server')
+      resources = @client.list_resources_for_type(feed_id, 'WildFly Server')
       wild_fly = resources[0]
 
       children = @client.list_child_resources(wild_fly, recursive: true)
 
-      expect(children.size).to be(211)
+      expect(children.size).to be(251)
```
abonas, line 150: why are the numbers fixed here (and below in several places) as part of the PR? what am I missing? :confused:

abonas:
I was trying to review this, but this PR is very big and also not separated to commits that might explain the changes separately. I left couple of comments, but I am definitely having difficulties understanding the code and the intention here. It might needs to be broken into smaller pieces. Or explain better in description what exactly is being done here.

spec/integration/inventory_spec.rb
jkremser: example

hawkularclient.gemspec
jkremser, line 30: because it uses the new feature from version `0.0.7` (https://github.com/Jiri-Kremser/simple-websocket-vcr/pull/6)

spec/integration/inventory_spec.rb
jkremser, line 150: So do you suggest doing something like 

``` ruby
expect(children.size).to be > x
```

instead?

As it is, the assumptions are stronger, it's rather question of taste what is better, I can argue that MiQ has the same thing for testing the # of features, that needs to be changed almost everytime and on other places there is something vague like `expect(something.size).to be >= 1`

But to be honest, I don't really care, I can change that to `> x`

jkremser:
Sorry for that, the PR isn't actually so big, most of the code is the VCR cassette templates. We can have a call if you want and I can describe the changes.

The main intention here is not to have PRs like this in the future ;) It's quite annoying to overwrite the VCR cassettes in the git repo. On the other hand, having the stale data isn't good too. Heiko started to using the ERB templating in the VCR. This way, you can 'parametrize' some data (this feature https://relishapp.com/vcr/vcr/v/3-0-1/docs/cassettes/dynamic-erb-cassettes).

This partially works. Issue with the dynamic cassettes is that there is no easy to keep them up to date with the real implementation. That's why I raised this PR, it still uses the ERB dynamic cassettes when running with no special evn vars, but at the same time, if running as `VCR_UPDATE=1 rspec`, it updates all the templates. The way I do that is different for VCR and for WebSocketVCR.

For WebSocketVCR it's a built in feature in the gem, that can be turned on by `reverse_substitution` option. As for the normal VCR, I have to do it manually here: https://github.com/hawkular/hawkular-client-ruby/pull/56/files#diff-93830fa29d616f7c87903d08b5b1b29aR107

Side note1: in Hawkular the feed id (basically the ID of the agent that collect the data from WildFly server) is a random UUID, so if you run the rspec against a live server, it'll be different all the time. Hence, the need for some parametrization of the data.

spec/integration/inventory_spec.rb
jkremser, line 52: You are right, it can be simplified to

``` ruby
    let(:cassette_name) do |e|
      e.description
    end
```

There was an issue with rubocop. It was complaining that it can be simplified even more as

``` ruby
let(:cassette_name, &:description)
```

=>

``` bash
spec/integration/inventory_spec.rb:50:25: C: Pass &:description as an argument to let instead of a block.
    let(:cassette_name) do |e|
                        ^^^^^^

25 files inspected, 1 offense detected
```

But if I did that, it didn't work. So this is kind of a dummy workaround to rubocop not to complain. To be honest, I don't know why the `let(:cassette_name, &:description)` is not working, if you have any better workaround, I am happy to change that.

CHANGES.rdoc
jkremser: It's easier to keep the track of the changes this way, but if we want to be super correct, it hasn't been released yet, so it should not be in the `CHANGES.rdoc`.

If you are asking about the 'All the clients have been unified into one', that is a left over from previous work and isn't logically related to this PR.

It's gone.

spec/integration/inventory_spec.rb
abonas: you might want to consider renaming it, as typically it is used for `exception` :)

spec/integration/inventory_spec.rb
abonas, line 52: What was the exact error/issue with the let(:cassette_name, &:description) ? it might also be rspec version related.
Regardless, you don't _have_ to follow all rubocop recommendations.

spec/integration/inventory_spec.rb
abonas, line 150: I didn't explain myself well. I thought that this PR's goal is to add support for erb templating as a general infra feature, so I was wondering why there was a need to update numbers in the specific tests.

abonas:
why is "Removing the 'not yet released' changes in CHANGES.rdoc" a separate commit?

abonas:
> Sorry for that, the PR isn't actually so big, most of the code is the VCR cassette templates. We can have a call if you want and I can describe the changes.
> 
> The main intention here is not to have PRs like this in the future ;) It's quite annoying to overwrite the VCR cassettes in the git repo. On the other hand, having the stale data isn't good too. Heiko started to using the ERB templating in the VCR. This way, you can 'parametrize' some data (this feature https://relishapp.com/vcr/vcr/v/3-0-1/docs/cassettes/dynamic-erb-cassettes).
> 
> This partially works. Issue with the dynamic cassettes is that there is no easy to keep them up to date with the real implementation. That's why I raised this PR, it still uses the ERB dynamic cassettes when running with no special evn vars, but at the same time, if running as VCR_UPDATE=1 rspec, it updates all the templates. The way I do that is different for VCR and for WebSocketVCR.
> 
> For WebSocketVCR it's a built in feature in the gem, that can be turned on by reverse_substitution option. As for the normal VCR, I have to do it manually here: https://github.com/hawkular/hawkular-client-ruby/pull/56/files#diff-93830fa29d616f7c87903d08b5b1b29aR107

`reverse_substitution` - is the name coming from somewhere built in? it is not very clear/how it relates to templates.

> Side note1: in Hawkular the feed id (basically the ID of the agent that collect the data from WildFly server) is a random UUID, so if you run the rspec against a live server, it'll be different all the time. Hence, the need for some parametrization of the data.

yes, the feature itself and its benefits are quite clear :)

jkremser:
> reverse_substitution - is the name coming from somewhere built in? it is not very clear/how it relates to templates.

It's an option for the `WebSocketVCR`, if it's used, it basically records/creates the templates. I.e. such cassettes that have variables on places where the values should be. Normal substitution goes the other way around, hence the name reverse substitution.

https://git.io/vw9RY
https://git.io/vw9RQ

jkremser:
> why is "Removing the 'not yet released' changes in CHANGES.rdoc" a separate commit?

Right, it doesn't make sense to introduce the change and then remove it withing the 1 PR, agreed. Can we just use the squash&merge thing on GitHub at the end? It's less work. (https://github.com/blog/2141-squash-your-commits)

abonas:
> Right, it doesn't make sense to introduce the change and then remove it withing the 1 PR, agreed. Can we just use the squash&merge thing on GitHub at the end? It's less work. (https://github.com/blog/2141-squash-your-commits)

I never used it this way, I'd typically squash or git amend manually. does it make sense to do it for a change that shouldn't have entered in the first place?

spec/integration/inventory_spec.rb
jkremser, line 150: Good question. Perhaps the server impl has been changed. Let me investigate it.

spec/integration/inventory_spec.rb
jkremser, line 52:
``` bash
Failures:

  1) Inventory Should list feeds
     Failure/Error: record('Inventory', @state, cassette_name, example: example)

     ArgumentError:
       no receiver given
     # ./spec/integration/inventory_spec.rb:57:in `block (2 levels) in <module:RSpec>'
...
25 failures
```

The `cassette_name` can't be used in the `around` interceptor.

I found less evil to have 

```
      description = example.description
      description
```

...in the code, rather than rubocop the very first consciously ignored offense :)

jkremser:
> does it make sense to do it for a change that shouldn't have entered in the first place?

Well, if you squash together 

```
+foo
```

and

```
-foo
```

there will be no `foo` in the resulting commit. But I've already done that, because I did also the `s/e/example/`

spec/integration/inventory_spec.rb
jkremser, line 150: yeah, the BTM module was added to Hawkular, that's why there are more resources

this is new: https://gist.github.com/Jiri-Kremser/591454a95330580249a5ae88d0a4fb4f
 (+ there are also some new JMS topics not in the gist)

abonas:
ok looking good :)

jkremser:
@pilhuhn could you please merge? If you rebase your pr on this, it'll fix the travis failures

PR #57, opened 2016-04-22
Resource hierarchy fixes

jkremser:
This is basically the same PR as the #55 but now targeting directly the master branch (not the 0.3.0 integration branch). It is based on PR #49 

It should fix the Issue #54 

All the methods dealing with resources expected the flat structure, but resources can be nested, so instead of `resource_id` as identifier we now use an array of those ids.

### DIFF
--- a/lib/hawkular/inventory/inventory_api.rb
+++ b/lib/hawkular/inventory/inventory_api.rb
@@ -56,29 +56,29 @@ module Hawkular::Inventory
     end
 
     # List resource types. If no need is given all types are listed
-    # @param [String] feed The id of the feed the type lives under. Can be nil for feedless types
+    # @param [String] feed_id The id of the feed the type lives under. Can be nil for feedless types
     # @return [Array<ResourceType>] List of types, that can be empty
-    def list_resource_types(feed = nil)
-      if feed.nil?
+    def list_resource_types(feed_id = nil)
+      if feed_id.nil?
         ret = http_get('/resourceTypes')
       else
-        the_feed = hawk_escape feed
+        the_feed = hawk_escape_id feed_id
         ret = http_get("/feeds/#{the_feed}/resourceTypes")
       end
       ret.map { |rt| ResourceType.new(rt) }
     end
 
     # Return all resources for a feed
-    # @param [String] feed Id of the feed that hosts the resources
+    # @param [String] feed_id Id of the feed that hosts the resources
     # @param [Boolean] fetch_properties Should the config data be fetched too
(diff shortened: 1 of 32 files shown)


lib/hawkular/inventory/inventory_api.rb
```diff
@@ -132,11 +138,11 @@ def get_config_data_for_resource(resource_id, feed)
     # @return [Array<Resource>] List of resources that are children of the given parent resource.
     #   Can be empty
     def list_child_resources(parent_resource, recursive = false)
```
pilhuhn, line 140: Shouldn't we go the full step here and require the canonical path `/t../f../r..` ?
I recall that I talked with @abonas some time ago that very often the actual objects are thrown away and only the "primary key" is retained. Users would still get the full object back from where they can `Resource.path` to get the path. 
While we are here, we could even make `Resource.id` the canonical path (primary key) and retain the original ìd` from Hawkular-Inventory under some different name.

lib/hawkular/inventory/inventory_api.rb
```diff
@@ -331,9 +337,22 @@ def create_resource_type(feed_id, type_id, type_name)
...
     def create_resource(feed_id, type_path, resource_id, resource_name = nil, properties = {})
-      the_feed = hawk_escape feed_id
+      create_resource_under_resource(feed_id, type_path, [], resource_id, resource_name, properties)
+    end
+
+    # Create a resource of a given type under a given feed. To retrieve that resource
+    # you need to call {#get_resource}
+    # @param [String] feed_id Id of the feed to add the resource to
+    # @param [String] type_path Path of the resource type of this resource
+    # @param [Array<String>] parent_resource_ids Ids of the resources under which we create this resource
```
pilhuhn, line 358: Same as above - instead of requiring an Array of "sub-ids", we should probably ask for the canonical path of the (to be) parent and do the computation internally if needed.

pilhuhn:
Besides the fact that this probably needs rebasing once #49 is merged, I think we should really re-think the concept of the `id`s and objects that we require as method parameters.
We would still return in all cases the full Resource/Metric/ResourceType object.

jkremser:
Why not, the API would be easier to use with the canonical paths, they are also stored in the MiQ db as ems_refs. But I'd not complicate this quite simple PR with it (the # of lines in this PR is misleading and it'll be less if the #49 is merged + rebased). 

I can do that in a subsequent one, purpose of this PR is to fix the issue with resource hierarchies following the same method signatures as in `0.2.2` version.

PR #59, opened 2016-05-03
Allow to inject events.

pilhuhn:
Allow to inject events from Ruby.

@jshaughn, @lucasponce  and @Jiri-Kremser please review
cc @abonas

### DIFF
--- a/lib/hawkular/alerts/alerts_api.rb
+++ b/lib/hawkular/alerts/alerts_api.rb
@@ -210,6 +210,22 @@ module Hawkular::Alerts
       query = generate_query_params(*criteria)
       http_get('/events' + query).map { |e| Event.new(e) }
     end
+
+    # Inject an event into Hawkular-alerts
+    # @param [String] id Id of the event must be unique
+    # @param [String] category Event category for further distinction
+    # @param [String] text Some text to the user
+    # @param [Hash<String,Object>] context Additional context information.
+    def create_event(id, category, text, context = {})
+      event = {}
+      event['id'] = id
+      event['ctime'] = Time.now.to_i * 1000
+      event['category'] = category
+      event['text'] = text
+      event['context'] = context unless context.nil?
+
+      http_post('/events', event)
+    end
   end
 
   # Representation of one Trigger
--- a/lib/hawkular/alerts/alerts_api.rb
+++ b/lib/hawkular/alerts/alerts_api.rb
@@ -222,7 +222,7 @@ module Hawkular::Alerts
       event['ctime'] = Time.now.to_i * 1000
       event['category'] = category
       event['text'] = text
-      event.merge!(extras) { |key, v1, v2| v1 }
+      event.merge!(extras) { |_key, v1, _v2| v1 }
 
       http_post('/events', event)
     end
--- a/lib/hawkular/alerts/alerts_api.rb
+++ b/lib/hawkular/alerts/alerts_api.rb
@@ -215,14 +215,14 @@ module Hawkular::Alerts
     # @param [String] id Id of the event must be unique
     # @param [String] category Event category for further distinction
     # @param [String] text Some text to the user
-    # @param [Hash<String,Object>] context Additional context information.
-    def create_event(id, category, text, context = {})
+    # @param [Hash<String,Object>] extras additional parameters
+    def create_event(id, category, text, extras)
       event = {}
       event['id'] = id
       event['ctime'] = Time.now.to_i * 1000
       event['category'] = category
       event['text'] = text
-      event['context'] = context unless context.nil?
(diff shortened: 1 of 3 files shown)


spec/integration/alerts_spec.rb
```diff
@@ -286,6 +286,13 @@ module Hawkular::Alerts::RSpec
       expect(events).to_not be_nil
       expect(events.size).to be(0)
     end
+
+    it 'Should create an event' do
```
lucasponce, line 290: Minor comment.
Perhaps a "describe" section for events ?
I'm also touching this file and to not mixed cassetes, splitting on "describe" could be nice.

lucasponce, line 290: Nevermind. After review, the describe Events is created.

lucasponce:
@pilhuhn It looks ok.

spec/integration/alerts_spec.rb
jkremser, line 290: The data in the cassette is stored in gzip form (https://github.com/pilhuhn/hawkular-client-ruby/blob/send_events/spec/vcr_cassettes/Alert/Events/Should_create_an_event.yml#L49) I guess, it's because the `vcr: { decode_compressed_response: true }` option is missing for the 'Alert/Events' (line 252)

pilhuhn, line 290: Do you know of a way to add this `decode_compressed_response: true` globally?

pilhuhn, line 290: I would like to leave it as is. And then in a followup PR we should change this for all of the Alert/.. categories that do not have it yet.

lib/hawkular/alerts/alerts_api.rb
```diff
@@ -210,6 +210,22 @@ def list_events(*criteria)
...
+    # @param [String] category Event category for further distinction
+    # @param [String] text Some text to the user
+    # @param [Hash<String,Object>] context Additional context information.
+    def create_event(id, category, text, context = {})
+      event = {}
+      event['id'] = id
+      event['ctime'] = Time.now.to_i * 1000
+      event['category'] = category
+      event['text'] = text
+      event['context'] = context unless context.nil?
```
abonas: in which case context will be nil? the default value is an empty hash

jshaughn: Hi Heiko, there are more fields that could be set.  Minimally I think you should add support for setting tags.  Perhaps the method should just take a hash as the last arg, like (id, category, text, optional_fields) and set whatever is supplied. That will keep the API from changing.

lib/hawkular/alerts/alerts_api.rb
```diff
@@ -210,6 +210,22 @@ def list_events(*criteria)
...
+
+      http_post('/events', event)
```
abonas, line 227: why are events needs to be injected via Ruby to hawkular side?

pilhuhn, line 227: On 3 May 2016, at 15:01, abonas wrote:

> why are events needs to be injected via Ruby to hawkular side?

There is no immediate need from the MiQ side, but as this is a general 
client gem,
other users may need it.

It certainly allows some testing of event catching in MiQ from within 
the Ruby code,
where the event is first injected into Hawkular and then after a bit of 
wait time
retrieved by the event catchers.

lib/hawkular/alerts/alerts_api.rb
```diff
@@ -210,6 +210,22 @@ def list_events(*criteria)
       query = generate_query_params(*criteria)
       http_get('/events' + query).map { |e| Event.new(e) }
     end
+
+    # Inject an event into Hawkular-alerts
+    # @param [String] id Id of the event must be unique
+    # @param [String] category Event category for further distinction
+    # @param [String] text Some text to the user
+    # @param [Hash<String,Object>] context Additional context information.
+    def create_event(id, category, text, context = {})
```
abonas: I'm pretty sure there's some syntactic sugar to create the keys more nicely in the hash from method parameters, but I can't remember it now :(

lib/hawkular/alerts/alerts_api.rb
pilhuhn: @abonas if someone passes in 'nil' explicitly
@jshaughn I could for sure just expect a hash or a `Event` object, but as the `Event` and others are `ostruct`, there is virtually no guidance in the gem-api on what is expected as input, but yes expecting some explicitly and then using the remainder from the hash would certainly work.

jshaughn: @pilhuhn I'm OK with any impl you choose as long as you extend the inject service to allow initial tags on the new event.  Without tags it is unlikely the event will be picked up by the event catcher given changes coming in the near future.

PR #60, opened 2016-05-04
Add support for group triggers

lucasponce:
Group triggers where not included in the preliminar mapping of the ruby client.
I have added partial support to cover the MiQ scenarios needed for the integration of MiQ Alerts and Hawkular Alerts.
So, this is not a fully 1-1 mapping between the Hawkular Alerts API in ruby, but extending the necessary API functions.
Please, @jshaughn, @pilhuhn, @Jiri-Kremser, @abonas take a look into the changes.
Thanks,

### DIFF
--- a/lib/hawkular/alerts/alerts_api.rb
+++ b/lib/hawkular/alerts/alerts_api.rb
@@ -360,7 +360,7 @@ module Hawkular::Alerts
     # Representing of one Condition
     class Condition
       attr_accessor :condition_id, :type, :operator, :threshold
-      attr_accessor :trigger_mode, :data_id
+      attr_accessor :trigger_mode, :data_id, :data2_id, :data2_multiplier
       attr_reader :condition_set_size, :condition_set_index, :trigger_id
 
       def initialize(cond_hash)
@@ -371,6 +371,8 @@ module Hawkular::Alerts
         @type = cond_hash['type']
         @trigger_mode = cond_hash['triggerMode']
         @data_id = cond_hash['dataId']
+        @data2_id = cond_hash['data2Id']
+        @data2_multiplier = cond_hash['data2Multiplier']
         @trigger_id = cond_hash['triggerId']
       end
 
@@ -383,6 +385,8 @@ module Hawkular::Alerts
         cond_hash['type'] = @type
         cond_hash['triggerMode'] = @trigger_mode
         cond_hash['dataId'] = @data_id
+        cond_hash['data2Id'] = @data2_id
+        cond_hash['data2Multiplier'] = @data2_multiplier
         cond_hash['triggerId'] = @trigger_id
         cond_hash
       end
--- a/lib/hawkular/alerts/alerts_api.rb
+++ b/lib/hawkular/alerts/alerts_api.rb
@@ -81,10 +81,93 @@ module Hawkular::Alerts
       http_post 'triggers/trigger', full_trigger
     end
 
+    # Creates the group trigger definition.
+    # @param [Trigger] trigger The group trigger to be created
+    # @return [Trigger] The newly created group trigger
+    def create_group_trigger(trigger)
+      ret = http_post 'triggers/groups', trigger.to_h
+      Trigger.new(ret)
+    end
+
+    # Updates a given group trigger definition
+    # @param [Trigger] trigger the group trigger to be updated
+    # @return [Trigger] The updated group trigger
+    def update_group_trigger(trigger)
+      http_put "triggers/groups/#{trigger.id}/", trigger.to_h
+      get_single_trigger trigger.id, false
+    end
+
+    # Creates the group conditions definitions.
+    # @param [String] trigger_id ID of the group trigger to set conditions
+    # @param [String] trigger_mode Mode of the trigger where conditions are attached (:FIRING, :AUTORESOLVE)
+    # @param [GroupConditionsInfo] group_conditions_info the conditions to set into the group trigger with the mapping
+    #                                                    with the data_id members map
+    # @return [Array<Condition>] conditions Array of associated conditions
+    def set_group_conditions(trigger_id, trigger_mode, group_conditions_info)
+      ret = http_put "triggers/groups/#{trigger_id}/conditions/#{trigger_mode}", group_conditions_info.to_h
+      conditions = []
+      ret.each { |c| conditions.push(Trigger::Condition.new(c)) }
+      conditions
+    end
+
+    # Creates a member trigger
+    # @param [GroupMemberInfo] group_member_info the group member to be added
+    # @return [Trigger] the newly created member trigger
+    def create_member_trigger(group_member_info)
+      ret = http_post 'triggers/groups/members', group_member_info.to_h
+      Trigger.new(ret)
+    end
+
+    # Detaches a member trigger from its group trigger
+    # @param [String] trigger_id ID of the member trigger to detach
+    # @return [Trigger] the orphan trigger
+    def orphan_member(trigger_id)
+      http_post "triggers/groups/members/#{trigger_id}/orphan", {}
+      get_single_trigger trigger_id, false
+    end
+
+    # Lists members of a group trigger
+    # @param [String] trigger_id ID of the group trigger to list members
+    # @param [boolean] orphans flag to include orphans
+    # @return [Array<Trigger>] Members found
+    def list_members(trigger_id, orphans = false)
+      ret = http_get "triggers/groups/#{trigger_id}/members?includeOrphans=#{orphans}"
+      ret.collect { |t| Trigger.new(t) }
+    end
+
+    # Creates a dampening for a group trigger
+    # @param [Dampening] dampening the dampening to create
+    # @return [Dampening] the newly created dampening
+    def create_group_dampening(dampening)
+      ret = http_post "triggers/groups/#{dampening.trigger_id}/dampenings", dampening.to_h
+      Trigger::Dampening.new(ret)
+    end
+
+    # Updates a dampening for a group trigger
+    # @param [Dampening] dampening the dampening to update
+    # @return [Dampening] the updated dampening
+    def update_group_dampening(dampening)
+      ret = http_put "triggers/groups/#{dampening.trigger_id}/dampenings/#{dampening.dampening_id}", dampening.to_h
+      Trigger::Dampening.new(ret)
+    end
+
+    # Deletes the dampening of a group trigger
+    # @param [String] trigger_id ID of the group trigger
+    # @param [String] dampening_id ID
+    def delete_group_dampening(trigger_id, dampening_id)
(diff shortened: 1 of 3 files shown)


lib/hawkular/alerts/alerts_api.rb
```diff
@@ -81,12 +81,100 @@ def create_trigger(trigger, conditions = [], dampenings = [], _actions = [])
       http_post 'triggers/trigger', full_trigger
     end
 
+    # Creates the group trigger definition.
+    # @param [Trigger] trigger The group trigger to be created
+    # @return [Trigger] The newly created group trigger
+    def create_group_trigger(trigger)
+      ret = http_post 'triggers/groups', trigger.to_h
```
abonas, line 88: no need for err handling? (e.g. when status is not 200 or 201) or it's handled elsewhere?

abonas, line 88: to_h and not to_json?

lib/hawkular/alerts/alerts_api.rb
```diff
@@ -81,12 +81,100 @@ def create_trigger(trigger, conditions = [], dampenings = [], _actions = [])
...
+      Trigger.new(ret)
+    end
+
+    # Updates a given group trigger definition
+    # @param [String] trigger_id ID of the group trigger to update
+    # @param [Trigger] trigger the group trigger to be updated
+    # @return [Trigger] The updated group trigger
+    def update_group_trigger(trigger_id, trigger)
```
abonas: why is there a need to pass both trigger id and trigger objects as parameters to this method and not extract the trigger id from the trigger object?

lib/hawkular/alerts/alerts_api.rb
```diff
@@ -81,12 +81,100 @@ def create_trigger(trigger, conditions = [], dampenings = [], _actions = [])
...
+    end
+
+    # Creates the group conditions definitions.
+    # @param [String] trigger_id ID of the group trigger to set conditions
+    # @param [String] trigger_mode Mode of the trigger where conditions are attached (:FIRING, :AUTORESOLVE)
+    # @param [GroupConditionsInfo] group_conditions_info the conditions to set into the group trigger with the mapping
+    #                                                    with the data_id members map
+    # @return [Array<Condition>] conditions Array of associated conditions
+    def set_group_conditions(trigger_id, trigger_mode, group_conditions_info)
+      trigger_mode = trigger_mode == :AUTORESOLVE ? 'autoresolve' : 'firing'
```
abonas: I'm not sure I understand the logic here regarding setting the trigger mode. shouldn't the relevant mode be passed as a parameter to the function and not set based on conditional logic inside?

lib/hawkular/alerts/alerts_api.rb
```diff
@@ -81,12 +81,100 @@ def create_trigger(trigger, conditions = [], dampenings = [], _actions = [])
...
+      ret = http_put "triggers/groups/#{trigger_id}/conditions/#{trigger_mode}", group_conditions_info.to_h
+      conditions = []
+      ret.each { |c| conditions.push(Trigger::Condition.new(c)) }
+      conditions
+    end
+
+    # Creates a member trigger
```
abonas, line 113: what/who is a member trigger? I am definitely not familiar with hawkular side, what confuses me here is the method name is create_group_member, but it returns object of type Trigger same as in create_group_trigger

lib/hawkular/alerts/alerts_api.rb
```diff
@@ -81,12 +81,100 @@ def create_trigger(trigger, conditions = [], dampenings = [], _actions = [])
...
+    def create_group_member(group_member_info)
+      ret = http_post 'triggers/groups/members', group_member_info.to_h
+      Trigger.new(ret)
+    end
+
+    # Detaches a member trigger from its group trigger
+    # @param [String] trigger_id ID of the member trigger to detach
+    # @return [Trigger] the orphan trigger
+    def orphan_member(trigger_id)
+      http_post "triggers/groups/members/#{trigger_id}/orphan", {}
```
abonas, line 125: clearly this is not an issue on the gem side, but why such operation is post and not update?

lib/hawkular/alerts/alerts_api.rb
```diff
@@ -81,12 +81,100 @@ def create_trigger(trigger, conditions = [], dampenings = [], _actions = [])
...
       http_delete "/triggers/#{trigger_id}"
     end
 
+    # Deletes the group trigger definition.
+    # @param [String] trigger_id ID of the group trigger to delete
+    def delete_group_trigger(trigger_id)
+      http_delete "/triggers/groups/#{trigger_id}"
+    end
+
+    # Deletes the dampening of a group trigger
```
abonas: perhaps worth to move it couple of lines up to be next to the other dampening methods? for readability only.

lib/hawkular/alerts/alerts_api.rb
lucasponce, line 88: Error handling is implemented at base_client.rb
I was just extending the API without touching major shared libraries, so perhaps in a major version error handling can be implemented in different way.
It happens the same with the to_h param, the json conversion is done at base_client.rb level, so the api expects to use the hash as input parameters for base http calls.

lib/hawkular/alerts/alerts_api.rb
jshaughn, line 113: It could be called create_member_trigger.  @abonas for some general overview on group triggers see http://www.hawkular.org/docs/dev/alerts.html#_group_triggers.  A group trigger is "template-level" and its member triggers are "instance-level".  In short, MIQ Alerts are very much like Group Triggers in H-Alerting.

lib/hawkular/alerts/alerts_api.rb
lucasponce: It could be. This method is mapping with the REST one defined at
https://github.com/hawkular/hawkular-alerts/blob/master/hawkular-alerts-rest/src/main/java/org/hawkular/alerts/rest/TriggersHandler.java#L488
so, I find it more interesting to maintain same parameter.

lib/hawkular/alerts/alerts_api.rb
jshaughn, line 125: Hmm, that is a good question, I'm not sure why the API has this as post.   Also, @lucasponce, is there really a need for orphans in the provider design, or is this just for making the gem more robust?

lib/hawkular/alerts/alerts_api.rb
lucasponce: +1
This was me, thinking on Java instead of ruby and strong typing I didn't realize that labels are converted to string.
Thanks for the catch.

lib/hawkular/alerts/alerts_api.rb
lucasponce, line 125: I think that in this POST method there was not a clear reason if should be implemented as POST or PUT. I guess that it could be implemented as a PUT as well. But this comes from the Hawkular Alerts.
@jshaughn in theory orphans won't be needed in the first iteration of MiQ provider but as I wasn't fully sure if we are going to need them in this task I prefer to add it on this PR.

lib/hawkular/alerts/alerts_api.rb
lucasponce: I wasn't unsure if we should group API by "group" or by dampening.
But anyway, not strong opinion on that.

lib/hawkular/alerts/alerts_api.rb
lucasponce, line 113: +1
I will change to create_member_trigger

lib/hawkular/alerts/alerts_api.rb
```diff
@@ -81,12 +81,100 @@ def create_trigger(trigger, conditions = [], dampenings = [], _actions = [])
...
+    def update_group_trigger(trigger_id, trigger)
+      http_put "triggers/groups/#{trigger_id}/", trigger.to_h
+      get_single_trigger trigger_id, false
+    end
+
+    # Creates the group conditions definitions.
+    # @param [String] trigger_id ID of the group trigger to set conditions
+    # @param [String] trigger_mode Mode of the trigger where conditions are attached (:FIRING, :AUTORESOLVE)
+    # @param [GroupConditionsInfo] group_conditions_info the conditions to set into the group trigger with the mapping
+    #                                                    with the data_id members map
```
jshaughn, line 104: Perhaps useful to add the notes: 
- The dataIdMemberMap should be null if the group has no members. 
- The dataIdMemberMap should be null if this is a DataDriven group trigger. In this case the member triggers are removed and will be re-populated as incoming data demands. 
- For [non-data-driven] group triggers with existing members the dataIdMemberMap is handled as follows. For members not included in the dataIdMemberMap their most recently supplied dataIdMap will be used. This means that it is not necessary to supply mappings if the new condition set uses only dataIds found in the old condition set. If the new conditions introduce new dataIds a full dataIdMemberMap must be supplied

lucasponce, line 104: I think these notes can go better into the GroupConditionsInfo class, instead of the method.

lib/hawkular/alerts/alerts_api.rb
lucasponce: @abonas you mean the Dampening class ?
I've reviewed the original API and there is not other dampening methods apart of the class.
As methods are grouped in one side and helpers classes in another I think we should maintain that at least for this iteration.
Perhaps in a major change of the gem we can change that.

abonas: no, I meant all dampening related methods in proximity to each other

lucasponce: ah ok, +1 I understood a different layout.
No problem, I will update that.

lib/hawkular/alerts/alerts_api.rb
pilhuhn: I also think the trigger id could/should be determined from the passed trigger object

lucasponce: Well, for me was more naturally to map it with the REST API, but anyway it is not a big deal to determine the id from the passed object.
 I can add that change to the PR.

lucasponce:
Is there a recommended way to use a gem PR into MiQ ?
(For example to validate that everything is correct and no missing functionality)
I guess that I can manually build it and copy files but I wonder if there is some way to point to this PR from a Gemfile or similar config.
Any suggestion is welcome.

abonas:
@lucasponce yes, for local checks you can modify miq Gemfile. specify the repository and branch to use for the gem (same line as the gem definition that already exists in gemfile). something like:
:git => 'git://github.com/user/yourrepo.git', :branch => 'mypatch'
or:
:github => 'user/yourrepo', :branch => 'mypatch'
you can also point it with the first example to your local git copy of the gem repo.

pilhuhn:
Thanks @lucasponce

PR #61, opened 2016-05-05
base client: adding support for proxy option

enoodle:


abonas:
imo the review comments from https://github.com/abonas/kubeclient/pull/169 stand here as well
update: at least the ones regarding tests, code clarity and readme.

enoodle:
I added a test in `spec/unit/base_client.rb` to test that the argument is passed to the rest_client that is created. The proxy functionality itself is implemented in RestClient and I assume that is tested there. I also changed the parameter in `options` to be named `http_proxy_uri` for clarity.

lib/hawkular/base_client.rb
```diff
@@ -26,6 +26,7 @@ def initialize(entrypoint = nil,
         verify_ssl: OpenSSL::SSL::VERIFY_PEER,
         ssl_client_cert: nil,
         ssl_client_key: nil,
+        http_proxy_uri: nil,
```
pilhuhn, line 29: 👍

Issue #52, opened 2016-04-17
Add functionality to list operations definitions in inventory


pilhuhn:
Allow to find operations/operation definitions for a resource (type?) in inventory

jkremser:
I'll add the list operations, however, not all the available operations are stored in the inventory. Currently, for WF there are: JDR, Reload, Deploy and Shutdown. But one can also, CRUD the deployments, JDBC datasources and drivers (these are not listed in inventory).

I was talking with Mazz about it and it's by design.

```
<mazz> anyone interacting with the HAwkular WildFly Agent will "just know" :) what it can support. This is our way of getting away from the pluggable agent we had in RHQ..
```

Oh, and there is also related JIRA on the inventory side: https://issues.jboss.org/browse/HWKINVENT-179

So, as far I understand it well, currently the operation types stored in the inventory are there only for the "security" reasosns, because every time, the agent executes the DMR operation, it also checks if it's white-listed in the inventory for given resource type.

pilhuhn:
I don't think this is right @jmazzitelli - why do we have Inventory when it is not used? 
There may be many agents or uses of the current Hawkular-agent in the future where we can't know in advance what operations are available or not. 
And then when you expect that someone writes a very specific UI per operation, you may be right, that the user needs to know the peculiarities of each operation. And then, the operation tag in the agent config is currently not self-explanatory either; I just recall <shutdown> where some additional data may be passed.

jmazzitelli:
The Hawkular WildFly Agent is not pluggable - so there is no way to add new websocket commands. The agent supports a fixed set of websocket command - these are defined here as JSON schemas: https://github.com/hawkular/hawkular-command-gateway/tree/master/hawkular-command-gateway-api/src/main/resources/schema

So this means the agent can support things like "Deploy Application" and "Add Datasource" and "Remove Datasource" and "Generate JDR Report". As we add new websocket commands, we'd get new websocket JSON schemas.

What you see in the standalone.xml metadata are DMR operations that are available on each DMR resource - like "Start", "Stop". These we store in inventory (without parameter metadata - just the operation names). If someone wants to invoke a DMR Operation, they pass over the websocket a JSON command ExecuteOperationRequest and pass in one of the operation names you need there (along with any parameters).

[...]

PR #63, opened 2016-05-09
Issue 52: adding a method for listing the operations stored in inventory

jkremser:
Only the method for retrieving all the operations declared in the inventory has been added. Inventory is storing more info under the `../resourceTypes/foo/operationTypes/bar/data` endpoint, but this endpoint is currently broken.

This fixes issue #52

### DIFF
--- a/lib/hawkular/inventory/inventory_api.rb
+++ b/lib/hawkular/inventory/inventory_api.rb
@@ -419,6 +419,28 @@ module Hawkular::Inventory
       MetricType.new(new_mt)
     end
 
+    # List operation definitions (types) for a given resource type
+    # @param [String] resource_type_path canonical path of the resource type entity
+    # @return [Array<String>] List of operation type ids
+    def list_operation_definitions(resource_type_path)
+      parsed_path = CanonicalPath.parse(resource_type_path.to_s)
+      feed_id = parsed_path.feed_id
+      resource_type_id = parsed_path.resource_type_id
+      ret = http_get("/feeds/#{feed_id}/resourceTypes/#{resource_type_id}/operationTypes")
+      ret.map { |ot| ot['id'] }
+    end
+
+    # List operation definitions (types) for a given resource
+    # @param [String] resource_path canonical path of the resource entity
+    # @return [Array<String>] List of operation type ids
+    def list_operation_definitions_for_resource(resource_path)
+      parsed_path = CanonicalPath.parse(resource_path.to_s)
+      feed_id = parsed_path.feed_id
+      resource_type_id = parsed_path.resource_type_id
+      ret = http_get("/feeds/#{feed_id}/resourceTypes/#{resource_type_id}/operationTypes")
+      ret.map { |ot| ot['id'] }
+    end
+
     # Create a Metric and associate it with a resource.
     # @param [String] metric_type_path Canonical path of the metric type of the new metric.
     # @param [String] resource_path Canonical path of the resource to which we want to associate the metric.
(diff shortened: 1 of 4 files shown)


jkremser:
@pilhuhn could you please review? It shouldn't be blocked by the fact that not all the operation types are in the inventory, right?

lib/hawkular/inventory/inventory_api.rb
```diff
@@ -429,6 +429,17 @@ def create_metric_type(feed_id, metric_type_id, type = 'GAUGE', unit = 'NONE', c
       MetricType.new(new_mt)
     end
 
+    # List operation types for given resource type
+    # @param [String] resource_type_path canonical path of the resource type entity
+    # @return [Array<String>] List of operation type ids
+    def list_operation_types(resource_type_path)
```
pilhuhn: Should we rename this to `list_operation_definitions` to make it more clear what it is - even if they are called differently in inventory?

pilhuhn: Should we (additionally) expose a `list_operation_definitions(resource_id)` which accepts a resource (CP) and then internally goes resource -> resourceType -> operationTypes?

pilhuhn:
In General LGTM. And no this missing stuff in inventory should not block this PR.

lib/hawkular/inventory/inventory_api.rb
```diff
@@ -429,6 +429,17 @@ def create_metric_type(feed_id, metric_type_id, type = 'GAUGE', unit = 'NONE', c
       MetricType.new(new_mt)
     end
 
+    # List operation types for given resource type
```
abonas: missing "a" - "for a given resource type"

lib/hawkular/inventory/inventory_api.rb
```diff
@@ -429,6 +429,17 @@ def create_metric_type(feed_id, metric_type_id, type = 'GAUGE', unit = 'NONE', c
...
+    # @param [String] resource_type_path canonical path of the resource type entity
+    # @return [Array<String>] List of operation type ids
```
abonas, line 424: the Array<String> is very Java, while the client is Ruby

lib/hawkular/inventory/inventory_api.rb
```diff
@@ -429,6 +429,17 @@ def create_metric_type(feed_id, metric_type_id, type = 'GAUGE', unit = 'NONE', c
...
+    def list_operation_types(resource_type_path)
+      parsed_path = CanonicalPath.parse(resource_type_path)
+      feed_id = parsed_path.feed_id
+      resource_type_id = parsed_path.resource_type_id
+      ret = http_get("/feeds/#{feed_id}/resourceTypes/#{resource_type_id}/operationTypes")
+      ret.map { |ot| ot['id'] }
```
abonas, line 430: why is it named operation type ids? (and not operation name/type)

jkremser, line 430: Because there is a conflict in the nomenclature, the `name` is reserved for the string that should be displayed to user, while id is some internal thing. Or at least that's how it was designed. Often, the 2 are the same though, it depends what agent reports to the inventory as `id` and `name`.

check https://git.io/vrqE5 to see how the data look like

lib/hawkular/inventory/inventory_api.rb
jkremser, line 424: I don't get your comment :] Array is normal data structure in Ruby http://docs.ruby-lang.org/en/2.0.0/Array.html  what's wrong with returning an array of stuff in Ruby?

lib/hawkular/inventory/inventory_api.rb
jkremser: Hmm, I can add that, but rather than overloading the method and doing some is_a? String ruby-fu, I'd create a new method called `list_operation_definitions_for_resource(resource_path)`, wdyt?

PR #66, opened 2016-05-10
Issue 64: Rewriting the inventory API to use the canonical path

### LINKED ISSUE #64: Rewrite the inventory client to use the canonical path whenever it's possible
A lot of methods takes the `feed_id` as argument, this can be obtained from the path.

here is the list:

```
list_resources_for_type(feed_id, type, fetch_properties = false)
->
list_resources_for_type(type_path, fetch_properties = false, filter = {})


get_config_data_for_resource(feed_id, res_ids)
->
get_config_data_for_resource(resource_path)


def list_child_resources(Resource parent_resource, recursive = false)
->
def list_child_resources(parent_resource_path, recursive = false)


list_relationships(Resource resource, named = nil)
->
list_relationships(path, named = nil)


list_relationships_for_feed(feed_id, named = nil)
->
list_relationships_for_feed(path, named = nil)


list_metrics_for_metric_type(feed_id, type)
->
list_metrics_for_metric_type(met_type_path)


list_metrics_for_resource_type(feed, type)
->
list_metrics_for_resource_type(res_type_path)


list_metrics_for_resource(Resource resource, filter = {})
->
list_metrics_for_resource(resource_path, filter = {})


create_resource(feed_id, type_path, resource_id, resource_name = nil, properties = {})
->
create_resource(type_path, resource_id, resource_name = nil, properties = {})


create_resource_under_resource(feed_id, type_path, parent_resource_ids, resource_id, resource_name = nil,
                                  properties = {})
->
create_resource_under_resource(type_path, parent_resource_path, resource_id, resource_name = nil,
                                  properties = {})

get_resource(feed_id, res_ids, fetch_resource_config = true)
->
get_resource(res_path, fetch_resource_config = true)


create_metric_for_resource(feed_id, metric_id, type_path, res_ids, metric_name = nil)
->
create_metric_for_resource(metric_type_path, resource_path, metric_id, metric_name = nil)


get_tenant
list_child_resources(parent_resource, recursive = false)
get_entity(path)
create_feed(feed_id, feed_name = nil)
delete_feed(feed_id)
create_resource_type(feed_id, type_id, type_name)
create_metric_type(feed_id, metric_type_id, type = 'GAUGE', unit = 'NONE', collection_interval = 60)
events(type = 'resource', action = 'created')
..untouched
```

jkremser:
fixes #64 

Part of this PR is also changing the `.rubocop.yml` making some rules weaker (`CyclomaticComplexity` and `PerceivedComplexity` both set to `9`)

### DIFF
--- a/spec/unit/canonical_path_spec.rb
+++ b/spec/unit/canonical_path_spec.rb
@@ -75,11 +75,11 @@ describe 'CanonicalPath' do
   end
 
   context 'with no tenant id' do
-    it 'should not be parseable' do
+    xit 'should not be parseable' do
       expect { CanonicalPath.parse('/f;myFeed/rt;resType') }.to raise_error(RuntimeError)
     end
 
-    it 'should not be constructed' do
+    xit 'should not be constructed' do
       expect { CanonicalPath.new(feed_id: 'something') }.to raise_error(RuntimeError)
     end
   end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,7 +5,10 @@
 # Overrides
 #
 CyclomaticComplexity:
+  Max: 9
   Severity: refactor
+PerceivedComplexity:
+  Max: 9
 Metrics/LineLength:
   Max: 120
 Metrics/ParameterLists:
--- a/lib/hawkular/base_client.rb
+++ b/lib/hawkular/base_client.rb
@@ -1,11 +1,14 @@
 require 'base64'
 require 'addressable/uri'
+require 'hawkular/hawkular_client_utils'
 
 module Hawkular
   # This is the base functionality for all the clients,
   # that inherit from it. You should not directly use it,
   # but through the more specialized clients.
   class BaseClient
+    include HawkularUtilsMixin
+
     # @!visibility private
     attr_reader :credentials, :entrypoint, :options
     # @return [Tenants] access tenants API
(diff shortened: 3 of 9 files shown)


jkremser:
@pilhuhn could you please review? 

I moved the `hawk_escape` function from `base_client.rb` to `HawkularUtilsMixin` because before those functions for encoding the ids were used internally by the api. Now, we can't do that internally, because it would end up with double encoded strings. All those funky characters are encoded in the canonical path, but if each method encoded it again it would end up with `/` -> `%2f` -> `%252F`

spec/unit/canonical_path_spec.rb
```diff
@@ -75,11 +75,11 @@
   end
 
   context 'with no tenant id' do
-    it 'should not be parseable' do
+    xit 'should not be parseable' do
```
pilhuhn, line 78: _x_it?

spec/vcr_cassettes/Inventory/Helpers/create_url.yml
```diff
@@ -44,5 +44,5 @@ http_interactions:
       encoding: UTF-8
       string: ''
     http_version: 
-  recorded_at: Thu, 05 May 2016 11:18:36 GMT
```
pilhuhn, line 47: Can you please not include this cassette?

pilhuhn:
I think that looks ok.

spec/unit/canonical_path_spec.rb
jkremser, line 78: `xit` disables the test

Before, I assumed it's a nonsense to create a canonical path without specifying a tenant. But now, all the methods in the inventory api that accept the canonical path, don't read the tenant id from it. It can be there, but don't have to. This test is testing if the creation of the `CanonicalPath` object fails if the tenant id is not specified.

oh an example:

``` ruby
path = CanonicalPath.new(feed_id: feed_id, resource_type_id: hawk_escape_id('WildFly Server'))
@client.list_resources_for_type(path.to_s, fetch_properties: false)
```

There is no need to specify the tenant id, because it's derived from the credentials in the inventory (this may change with the tenant header)

spec/vcr_cassettes/Inventory/Helpers/create_url.yml
jkremser, line 47: Right, this is not necessary for the PR.

jkremser, line 47: I've removed also the `spec/vcr_cassettes/Inventory/Helpers/get_feeds.yml` cassette from the commit

PR #67, opened 2016-05-11
Let the gem always provide HTTP header Hawkular-Tenant

jshaughn:
Use default value 'hawkular'. It can still be overidden by the client options.

@Jiri-Kremser please take a look, this logic may not be right but hopefully it's just simple like this.
@pilhuhn take a look and see whether you think this is in line with our decision to just send the default header value.

### DIFF
--- a/lib/hawkular/base_client.rb
+++ b/lib/hawkular/base_client.rb
@@ -182,8 +182,8 @@ module Hawkular
     end
 
     def tenant_header
-      @options[:tenant].nil? ? {} : { :'Hawkular-Tenant' => @options[:tenant],
-                                      'tenantId' => @options[:tenant] }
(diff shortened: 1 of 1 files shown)


pilhuhn:
Rubocop is unhappy https://travis-ci.org/hawkular/hawkular-client-ruby/jobs/129422639#L218
otherwise I think this is good.

jkremser:
I can see the headers in the yml cassettes if I run `VCR_UPDATE=1 rspec` against live hawkular server:

```
...
       - application/json
       Accept-Encoding:
       - gzip, deflate
+      Hawkular-Tenant:
+      - hawkular
       Content-Type:
       - application/json
       User-Agent:
...
```

lgtm

PR #70, opened 2016-05-13
Simple CLI

jkremser:
Using the IRB + some customization as a simple CLI client


hawkular_client_cli.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+IRBRC=.irbrc.rb irb
```
pilhuhn, line 2: Are you sure this is complete? Shouldn't it call irb to to anything?

abonas:
I think missing documentation usage examples as part of such PRs is problematic.
How can someone understand how to use the gem? (actually, even without the cli)

Also, please explain what is the motivation of this PR.
Here are some points I wonder about:
1. This gem is already a ruby gem. which implies its users have ruby installed and can open irb or rails console on their own and use the gem. So who is the target audience of the cli in this case? If the motivation is for non ruby users to use hawkular, then I am wondering why this should be part of this gem. (it could be a separate library that uses the gem, or part of any other python/java/etc. client too, and then there also could be duplicity with this one if someone else decides to add another cli client)
2. Fixing bugs in the cli related code will require gem release and versioning for the entire gem, incompatible changes and fixes on the cli part will require bumping the version - that will have its implications.

PR #72, opened 2016-05-27
Adding a test for SSL communication w/o verifying the certificate. Al…

jkremser:
…so making the clients to take into account the options hash - this isn't a breaking change.

### DIFF
--- a/spec/integration/inventory_spec.rb
+++ b/spec/integration/inventory_spec.rb
@@ -99,6 +99,21 @@ module Hawkular::Inventory::RSpec
       expect(resources.size).to be(2)
     end
 
+    it 'Should list feeds when using SSL without certificate' do
+      # change this to the real credentials when updating the VCR
+      @state[:super_secret_username] = 'username'
+      @state[:super_secret_password] = 'password'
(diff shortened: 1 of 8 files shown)


abonas:
I suggest creating some documentation for the gem users where this option is mentioned.

jkremser:
I need this to be merged in so that the options propagation among clients work ok (possibly w/ the tenant)

spec/integration/inventory_spec.rb
```diff
@@ -99,6 +99,21 @@ module Hawkular::Inventory::RSpec
       expect(resources.size).to be(2)
     end
 
+    it 'Should list feeds when using SSL without certificate' do
+      # change this to the real credentials when updating the VCR
+      @state[:super_secret_username] = 'username'
+      @state[:super_secret_password] = 'password'
+      creds = { username: @state[:super_secret_username],
+                password: @state[:super_secret_password] }
+      tori_url = 'https://hawkular.torii.gva.redhat.com/hawkular/inventory'
```
abonas, line 108: a unitest should not include references to such environments, especially if not everyone have access to them.

PR #73, opened 2016-06-13
Fix list resources for feed bailing out on unknown variable

pilhuhn:
Api calls to `inventory_api#list_resources_for_feed` fail with 

```
NameError: undefined local variable or method `feed' for #<Hawkular::Inventory::InventoryClient:0x007fbe2acd1708>
./lib/hawkular/inventory/inventory_api.rb:76:in `list_resources_for_feed'
```


abonas:
LGTM

PR #74, opened 2016-06-15
Adding a way to omit the Hawkular-Tenant header.

jkremser:
empty string disables it. It's better than using the nil value, because both

``` ruby
{}[:tenant].nil?
{:tenant=>nil}[:tenant].nil?
```

..return true

### DIFF
--- a/lib/hawkular/base_client.rb
+++ b/lib/hawkular/base_client.rb
@@ -146,8 +146,14 @@ module Hawkular
     end
 
     def tenant_header
-      @options[:tenant].nil? ? { 'Hawkular-Tenant' => 'hawkular' } : { :'Hawkular-Tenant' => @options[:tenant],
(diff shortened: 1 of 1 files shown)


lib/hawkular/base_client.rb
```diff
@@ -146,8 +146,14 @@ def token_header
     end
 
     def tenant_header
-      @options[:tenant].nil? ? { 'Hawkular-Tenant' => 'hawkular' } : { :'Hawkular-Tenant' => @options[:tenant],
-                                                                       'tenantId'         => @options[:tenant] }
+      if @options[:tenant].nil?
+        { 'Hawkular-Tenant' => 'hawkular' }
```
abonas, line 150: shouldn't be the default. user should pass the needed tenant (similar to that you won't set user/password default values)

PR #75, opened 2016-06-15
If the tenant is not passed in the options no default value is used.

jkremser:
cc @abonas

### DIFF
--- a/lib/hawkular/base_client.rb
+++ b/lib/hawkular/base_client.rb
@@ -146,14 +146,9 @@ module Hawkular
     end
 
     def tenant_header
-      if @options[:tenant].nil?
-        { 'Hawkular-Tenant' => 'hawkular' }
-      elsif @options[:tenant].empty?
-        {}
-      else
-        { :'Hawkular-Tenant' => @options[:tenant],
-          'tenantId'         => @options[:tenant] }
-      end
+      headers = {}
+      headers[:'Hawkular-Tenant'] = @options[:tenant] unless @options[:tenant].nil?
+      headers
     end
 
     def handle_fault(f)


lib/hawkular/base_client.rb
```diff
@@ -146,14 +146,8 @@ def token_header
...
-      if @options[:tenant].nil?
-        { 'Hawkular-Tenant' => 'hawkular' }
-      elsif @options[:tenant].empty?
-        {}
-      else
-        { :'Hawkular-Tenant' => @options[:tenant],
-          'tenantId'         => @options[:tenant] }
-      end
+      @options[:tenant].nil? ? {} : { :'Hawkular-Tenant' => @options[:tenant],
+                                      'tenantId'         => @options[:tenant] }
```
jpkrohling: Is there a place that documents where this `tenantId` is used? Also, why is the first a symbol and the second a string?

abonas: isn't here the same information sent twice (2 keys with same value)? what am I missing?

jkremser:
this PR is basically reverting the state to the original state before PR (https://github.com/hawkular/hawkular-client-ruby/pull/67)

I am not the author of the original code => no idea why 'tenantId' is there twice, but I left it there to be safe.

> Is there a place that documents where this tenantId is used?

I am not aware of it, but it'd be useful to have something. @jpkrohling do you think it's safe to remove the second occurrence? `'Hawkular-Tenant'` is 100% correct, but I am not aware of any component that needs also `'tenantId'` header.

jpkrohling:
It seems it's there since the first version from Libor: https://github.com/hawkular/hawkular-client-ruby/blob/73848c5f611ea31675483f9d92a10a57ee64cb41/lib/hawkularclient.rb#L107 . It says to remove it once HWKMETRICS-187 is fixed , which seems to be the case.

jkremser:
> It seems it's there since the first version from Libor

I knew it's either Libor's or vrockai's fault :)

pilhuhn:
As HWKMETRICS-187 says it was fixed in 0.7, so with metrics 0.7+ we are indeed save to completely leave out the 'tenantId' header

PR #76, opened 2016-06-20
Changes + version bump (2.0.0)

jkremser:
big TODO: tests


abonas:
if the tenant related functionality has changed comparing to previous gem version (or for that matter, any other functionality), then the version should be bumped differently (major)

PR #77, opened 2016-06-21
Add status endpoint for Alerting client

lucasponce:
Minor PR to add status endpoint into Alerting API.

### DIFF
--- a/lib/hawkular/alerts/alerts_api.rb
+++ b/lib/hawkular/alerts/alerts_api.rb
@@ -20,6 +20,13 @@ module Hawkular::Alerts
       super(entrypoint, credentials, options)
     end
 
+    # Return version and status information for the used version of Hawkular-Alerting
+    # @return [Hash{String=>String}]
+    #         ('Implementation-Version', 'Built-From-Git-SHA1', 'status')
+    def fetch_version_and_status
+      http_get('/status')
+    end
+
     # Lists defined triggers in the system
     # @param [Array] ids List of trigger ids. If provided, limits to the given triggers
     # @param [Array] tags List of tags. If provided, limits to the given tags. Individual
--- a/spec/integration/alerts_spec.rb
+++ b/spec/integration/alerts_spec.rb
@@ -424,13 +424,16 @@ module Hawkular::Alerts::RSpec
     end
   end
 
-  # TODO: enable when alerts supports it
-  # describe 'Alerts' do
-  #   it 'Should return the version' do
-  #     data = @client.get_version_and_status
-  #     expect(data).not_to be_nil
-  #   end
-  # end
+  describe 'Alerts', vcr: { decode_compressed_response: true } do
+    before(:each) do
+      @client = Hawkular::Alerts::AlertsClient.new(ALERTS_BASE, creds)
+    end
+
+    it 'Should return the version' do
+      data = @client.fetch_version_and_status
+      expect(data).not_to be_nil
+    end
+  end
 
   describe 'Alert/Events', :vcr do
     VCR.configure do |c|
--- a/spec/vcr_cassettes/Alerts/Should_return_the_version.yml
+++ b/spec/vcr_cassettes/Alerts/Should_return_the_version.yml
@@ -1,0 +1,40 @@
+---
+http_interactions:
+- request:
+    method: get
+    uri: http://jdoe:password@localhost:8080/hawkular/alerts/status
+    body:
+      encoding: US-ASCII
+      string: ''
+    headers:
+      Accept:
+      - application/json
+      Accept-Encoding:
+      - gzip, deflate
+      Content-Type:
+      - application/json
+      User-Agent:
+      - Ruby
+  response:
+    status:
+      code: 200
+      message: OK
+    headers:
+      Connection:
+      - keep-alive
+      X-Powered-By:
+      - Undertow/1
+      Server:
+      - WildFly/10
+      Content-Type:
+      - application/json
+      Content-Length:
+      - '124'
+      Date:
+      - Tue, 21 Jun 2016 07:58:51 GMT
+    body:
+      encoding: UTF-8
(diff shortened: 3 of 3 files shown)


lib/hawkular/alerts/alerts_api.rb
```diff
@@ -20,6 +20,13 @@ def initialize(entrypoint = 'http://localhost:8080/hawkular/alerts', credentials
       super(entrypoint, credentials, options)
     end
 
+    # Return version and status information for the used version of Hawkular-Alerting
+    # @return [Hash{String=>String}]
+    #         ('Implementation-Version', 'Built-From-Git-SHA1', 'status')
+    def fetch_version_and_status
```
abonas, line 26: I suggest to shorten it to `version_and_status`
I also wonder (definitely not relevant to this patch) why the version/built from keys are so long?
'version' and 'git-sha' could be nicer to read and shorter to use.

lucasponce, line 26: def fetch_version_and_status is the API method used in metrics.

abonas, line 26: @lucasponce it doesn't mean it's a good thing :wink: 
I think it's really too long for the client users.

lucasponce, line 26: This PR is not for changing the API.
It is just to include a missing API into the same consistency naming used before.
The name of the method is not important here, it can be longer or shorter, the point is to add it into a consistency way.
A change of the API can follow this one.

abonas, line 26: It is clear that the api is not changed by this PR.
However I do think that when new code that is added - and this PR does add a new method for the client users, it can and should be done better than the past if possible. Why add something if you already know you'd like to change it after? Otherwise the client will be always compliant with the past and it will be really hard to improve it. It's not a critical issue, just my 2 cents.

lucasponce, line 26: Sorry, I am not agree here.
def fetch_version_and_status is not new code, it is to complete the same method used in other subclient to another subclient.
A change on the method will imply a bump of the version, and for this minor version I don't think is worth to do it.
So, a review of the PR (and its discussion) should go in a following PR not in this.

abonas, line 26: Perhaps I am missing something here - aren't you adding a new method to this gem in this PR?

pilhuhn, line 26: I don't have a strong opinion here, but can seen Lukas' point of consistency. If you know the method form metrics, you know it from alerts.

lucasponce, line 26: Yes and no.
Yes, at it is not covered in alerts subclient how to fetch the version.
No, as it is already part of the public api for metrics subclient:
https://github.com/hawkular/hawkular-client-ruby/blob/master/lib/hawkular/metrics/metric_api.rb#L18
So, it has not sense to have different names for the same function in different subclients.
And this PR has not intention to review the API in all places where this method is used.
As I commented, this PR is just to complete to add this feature for future uses.

abonas, line 26: Yes, but if we agree that the method name is nicer in its shorter version (and it's definitely 'if' since it's a matter of taste), doesn't it make more sense to change later the one in metrics, instead of changing later both of them? (with the implication of fixing more places that call those methods). plus, as you know, not everyone who uses metrics uses alerts.

abonas, line 26: as I said, this is _not_ a critical issue, but I do think we should start looking how usable and convenient is the client, and the best way to start is putting an effort to make it like that through new code.

lucasponce, line 26: My opinion is that change should go in  a following PR where we can review and upgrade version later.

abonas, line 26: it's definitely up to the gem maintainers.

jkremser, line 26: I am ok with `fetch_version_and_status`. Perhaps, it's not an idiomatic Ruby, but at least it says, it'll actually does something (fetch -> do a rest call).

I assume these methods can be used in the future versions of gem, when the different versions of the h-services REST API will require different treatment by the client. Potentially frequently (depends on the impl). To do that, we can actually store the version and/or sha as a client attributes, perhaps somehow lazily (do the actual fetch, if it's firstly asked for them).

So what about having `fetch_version_and_status` for each client and + the top-lvl client that can call it on the all sub-clients and returns the Hash of Hash of versions. And then `version` field on all the clients that will represent a cached value of the version + does the lazy fetch logic. wdyt?

This way, this PR can be merged and the `fetch_version_and_status` will be on all the clients (actually, it still needs to be added to the inventory client) => consistency.
And we can later add the `version` or perhaps `backend_version` fields that will expose the cached value => adding new stuff won't break the API

abonas, line 26: > I am ok with fetch_version_and_status. Perhaps, it's not an idiomatic Ruby, but at least it says, it'll actually does something (fetch -> do a rest call).

the majority of the methods in client are actually doing a call, their names are not implying that and it's an implementation detail.

> I assume these methods can be used in the future versions of gem, when the different versions of the h-services REST API will require different treatment by the client. Potentially frequently (depends on the impl). To do that, we can actually store the version and/or sha as a client attributes, perhaps somehow lazily (do the actual fetch, if it's firstly asked for them).
> 
> So what about having fetch_version_and_status for each client and + the top-lvl client that can call it on the all sub-clients and returns the Hash of Hash of versions. And then version field on all the clients that will represent a cached value of the version + does the lazy fetch logic. wdyt?

imo caching is not needed for such a minor thing

> This way, this PR can be merged and the fetch_version_and_status will be on all the clients (actually, it still needs to be added to the inventory client) => consistency.
> And we can later add the version or perhaps backend_version fields that will expose the cached value => adding new stuff won't break the API

it can be merged imo, it's a minor addition

PR #83, opened 2016-06-27
Version bump and release notes for V2.1.0

pilhuhn:
Prepare for a 2.0.1 version
- Operations bugfixes
- Alerts /status endpoint

cc @abonas @Jiri-Kremser @lucasponce


abonas:
it should be 2.1.0 because a new method was added per [item 2 here](http://semver.org/)

PR #87, opened 2016-06-28
Fix check of enabled flag for add_deployment operation

### LINKED ISSUE #86: Add deployment operation always sets enabled to true
When adding a deployment it is always set to be enabled, even if the enabled field is set to false.

ammendonca:
The enabled flag was being set with `hash[:enabled] ||= true` which, if enabled
would be defined and set to false, it would evaluate as false and set it to true.

This checks if it is defined and only if it is not, it sets the default to true.

Fixes #86

### DIFF
--- a/lib/hawkular/operations/operations_api.rb
+++ b/lib/hawkular/operations/operations_api.rb
@@ -134,7 +134,7 @@ module Hawkular::Operations
     #
     # @param callback [Block] callback that is run after the operation is done
     def add_deployment(hash, &callback)
-      hash[:enabled] ||= true
+      hash[:enabled] = hash.key?(:enabled) ? hash[:enabled] : true
       required = [:resource_path, :destination_file_name, :binary_content]
       check_pre_conditions hash, required, &callback
 


lib/hawkular/operations/operations_api.rb
```diff
@@ -134,7 +134,7 @@ def invoke_specific_operation(operation_payload, operation_name, &callback)
     #
     # @param callback [Block] callback that is run after the operation is done
     def add_deployment(hash, &callback)
-      hash[:enabled] ||= true
+      hash[:enabled] = defined?(hash[:enabled]) ? hash[:enabled] : true
```
abonas: not sure what does the `defined?` return in this case, but why not using `hash[:enabled].key?`

ammendonca: The result is the same, but indeed it seems more adequate to use the specific hash method `key?` instead of the generic `defined?`. Will change.

pilhuhn:
I wonder if this enabled/disabled should also be checked in 'Operations/Add deployment should be doable' ?

PR #88, opened 2016-06-29
Add method on Metrics API for fetching raw data for multiple metrics.…

josejulio:
… (HWKMETRICS-393)

### DIFF
--- a/lib/hawkular/metrics/metric_api.rb
+++ b/lib/hawkular/metrics/metric_api.rb
@@ -129,6 +129,18 @@ module Hawkular::Metrics
         resp.is_a?(Array) ? resp : [] # API returns no content (empty Hash) instead of empty array
       end
 
+      # Retrieve raw data for multiple metrics.
+      # @param ids [Array[String]] metric definition ids
+      # @param starts [Integer] optional timestamp (default now - 8h)
+      # @param ends [Integer] optional timestamp (default now)
+      # @param limit [Integer] optional limit the number of data points returned
+      # @param order [String] optional Data point sort order, based on timestamp (ASC, DESC)
+      # @return [Array[Hash]] named datapoints
+      def raw_data(ids, starts: nil, ends: nil, limit: nil, order: nil)
+        params = { ids: ids, start: starts, end: ends, limit: limit, order: order }
+        @client.http_post("/#{@resource}/raw/query", params)
+      end
+
       # Retrieve metric datapoints by tags
       # @param tags [Hash]
       # @param starts [Integer] optional timestamp (default now - 8h)
--- a/spec/integration/metric_spec.rb
+++ b/spec/integration/metric_spec.rb
@@ -70,6 +70,61 @@ describe 'Mixed metrics' do
   end
   #  end
 
+  it 'Should requests raw data for multiple metrics' do
+    @client = setup_client(username: 'jdoe', password: 'password', tenant: 'vcr-test')
+    ids = [SecureRandom.uuid, SecureRandom.uuid, SecureRandom.uuid]
+    VCR.use_cassette('Mixed_metrics/Should requests raw data for multiple metrics',
+                     erb: { ids: ids }, record: :none
+                    ) do
+      expect(@client.counters.raw_data(ids).size).to be 0
+      expect(@client.gauges.raw_data(ids).size).to be 0
+      expect(@client.avail.raw_data(ids).size).to be 0
+
+      @client.push_data(
+        counters: [
+          { id: ids[0], data: [{ value: 1 }] },
+          { id: ids[1], data: [{ value: 2 }] },
+          { id: ids[2], data: [{ value: 3 }] }
+        ],
+        availabilities: [
+          { id: ids[0], data: [{ value: 'up' }] },
+          { id: ids[1], data: [{ value: 'down' }] },
+          { id: ids[2], data: [{ value: 'up' }] }
+        ],
+        gauges: [
+          { id: ids[0], data: [{ value: 1.1 }] },
+          { id: ids[1], data: [{ value: 2.2 }] },
+          { id: ids[2], data: [{ value: 3.3 }] }
+        ]
+      )
+
+      counter_metrics = @client.counters.raw_data(ids)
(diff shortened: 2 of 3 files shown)


lib/hawkular/metrics/metric_api.rb
```diff
@@ -129,6 +129,19 @@ def get_data(id, starts: nil, ends: nil, bucketDuration: nil, buckets: nil, perc
...
 
+      # Retrieve raw data for multiple metrics.
+      # This API is experimental and may undergo non-backwards compatible changes on the future
+      # @param ids [Array[String]] metric definition ids
+      # @param starts [Integer] optional timestamp (default now - 8h)
+      # @param ends [Integer] optional timestamp (default now)
+      # @param limit [Integer] optional limit the number of data points returned
+      # @param order [String] optional Data point sort order, based on timestamp (ASC, DESC)
+      # @return [Array[Hash]] named datapoints
+      def raw_data(ids, starts: nil, ends: nil, limit: nil, order: nil)
```
abonas, line 139: a more clear name would be `metric_ids` instead of just `ids`
regarding limit - perhaps max_data_points?

lib/hawkular/metrics/metric_api.rb
```diff
@@ -129,6 +129,19 @@ def get_data(id, starts: nil, ends: nil, bucketDuration: nil, buckets: nil, perc
...
+        params = { ids: ids, start: starts, end: ends, limit: limit, order: order }
+        @client.http_post("/#{@resource}/raw/query", params)
```
abonas, line 141: @pilhuhn I wonder why retrieval of metrics should be done in POST rather than GET...

pilhuhn, line 141: Believe me, you don't want to know that :-)

Seriously: metric ids can be very long (strings with crazy characters) and querying several of them at once can lead to very long and complicated query parts of the GET requests, so the hawkular-metrics team decided to deliver the metric ids as payload of a POST request.
I don't think it is that bad, because the caching argument (GET vs POST) is not really applicable for metric data, as different requests usually ask for different time ranges and return different data each time.

lib/hawkular/metrics/metric_api.rb
```diff
@@ -129,6 +129,19 @@ def get_data(id, starts: nil, ends: nil, bucketDuration: nil, buckets: nil, perc
         resp.is_a?(Array) ? resp : [] # API returns no content (empty Hash) instead of empty array
       end
 
+      # Retrieve raw data for multiple metrics.
+      # This API is experimental and may undergo non-backwards compatible changes on the future
```
pilhuhn: This scares me. We should try to get it right from the start.

spec/integration/metric_spec.rb
```diff
@@ -70,6 +70,25 @@
...
+  it 'Should requests raw data for multiple metrics' do
+    @client = setup_client(username: 'jdoe', password: 'password')
+    ids = [SecureRandom.uuid, SecureRandom.uuid, SecureRandom.uuid]
+    VCR.use_cassette('Mixed_metrics/Should requests raw data for multiple metrics',
+                     erb: { ids: ids }, record: :none
+                    ) do
+      expect(@client.avail.raw_data(ids).size).to be 0
+
+      @client.push_data(
+        availabilities: [
```
pilhuhn, line 89: Can you please also test for a mix of availability, gauge and counter?

spec/integration/metric_spec.rb
```diff
@@ -70,6 +70,25 @@
...
+          { id: ids[0], data: [{ value: 'up' }] },
+          { id: ids[1], data: [{ value: 'down' }] },
+          { id: ids[2], data: [{ value: 'up' }] }
+        ]
+      )
+      expect(@client.avail.raw_data(ids).size).to be 3
```
pilhuhn: Can you please explicitly test that the right datapoints in the expected format are returned?

lib/hawkular/metrics/metric_api.rb
abonas, line 141: ok thanks, now I know :)

lib/hawkular/metrics/metric_api.rb
josejulio, line 139: Being on the file metric_api and Metrics class, should we be verbose enough to refer to it as metric_ids ? 
I don't have any problem on changing them. I was just trying to mimic the params on get_data[1] method. Almost the same params can be used on both methods and I don't see why we should name 'limit' differently on this method as this could lead to confusions.

Should I change `ids` to `metric_ids` and `limit` to `max_data_points`?

[1] https://github.com/josejulio/hawkular-client-ruby/blob/2fe78d22f13cf45cf611a4e9cf859eb49575c921/lib/hawkular/metrics/metric_api.rb#L123

lib/hawkular/metrics/metric_api.rb
josejulio: I'm just passing the warning that appears on the hawkular-metrics[1]. Should I remove?

[1] https://github.com/jsanda/hawkular-metrics/blob/e559e28482f51dd30805b63da09efce71538490b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/CounterHandler.java#L268-L269

spec/integration/metric_spec.rb
josejulio, line 89: Sure

spec/integration/metric_spec.rb
josejulio: Missed that, will do

josejulio:
Rebased to master and now it passes the tests.

lib/hawkular/metrics/metric_api.rb
pilhuhn, line 139: I have no strong feelings about ids/metric_ids, but I guess the latter is more clear. Or something like 'id_list'.

For `limit` I think it makes sense to leave as is, as this is what the rest-api docs name them too. See e.g. https://github.com/hawkular/hawkular-metrics/blob/master/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/AvailabilityHandler.java#L298

PR #89, opened 2016-06-29
Updating VCRs (metrics)

jkremser:
- making it possible to test both old metrics `0.8.0` (the "OpenShift" ones) and metrics that are part of the hawkular-services (probably only the latest)


spec/integration/metric_spec.rb
```diff
@@ -7,489 +7,524 @@
...
+    puts 'skipping v8 metrics'
+    next
   end
-
-  it 'Should create and return tenant' do
-    tenant = SecureRandom.uuid
-    @client.tenants.create(tenant)
-    created = @client.tenants.query.select { |t| t.id == tenant }
-    expect(created).not_to be nil
+  if ENV['SKIP_SERVICES_METRICS'] == '1' && metrics_context == services_context
```
lucasponce, line 19: Perhaps these variables should be also documented on the README.rdoc.

spec/integration/metric_spec.rb
```diff
@@ -7,489 +7,524 @@
...
-      data = @client.gauges.get_data(id)
-      expect(data.size).to be 1
+    around(:each) do |example|
+      run_for = example.metadata[:run_for]
+      if run_for.nil? || run_for.empty? || run_for.include?(metrics_context)
+        @random_id = SecureRandom.uuid
+        if example.metadata[:skip_auto_vcr]
+          example.run
+        else
+          record("Metrics/#{metrics_context}", { id: @random_id }, cassette_name, example: example)
```
lucasponce, line 40: Just to get familiar, I have used VCR in a basic level.
record is part of the VCR api or it is a helper done for the spec ?

lucasponce:
I ran locally the PR and I have some pending tests:
https://paste.fedoraproject.org/388702/14679065/

I guess that is expected, or did I miss anything on my command line ?

jkremser:
> I guess that is expected, or did I miss anything on my command line ?

Yes, that's expected. Here is an example of a test that is skipped for v8 context, because it tests a feature that hasn't been there
https://git.io/vKmra

spec/integration/metric_spec.rb
jkremser, line 19: +1, done

spec/integration/metric_spec.rb
jkremser, line 40: a helper (https://git.io/vKmoW)

Good thing about it, is that the old tests had to repeat the name and all the previous contexts in the string that was used for `VCR.use_cassette` (like here: https://git.io/vKmoQ), this way, it's auto-inferred + the templating works

lucasponce:
In general, after reviewing closely the code LGTM.
I have a comment, but it is not a blocker from my side.
In general the separation of templates in the folder:
spec/vcr_cassettes/Metrics/metrics_0_8_0
spec/vcr_cassettes/Metrics/metrics_services
It is good, but I would like to have also the version of the services.
I mean something like:
spec/vcr_cassettes/Metrics/metrics_services_0_5_0
In that reason, it is helpful for the future to know against which version the cassettes were recorded.
This can be done in a following PR, but I guess it will help to spot potential issues due version changes.

jkremser:
> It is good, but I would like to have also the version of the services ... This can be done in a following PR

+100

PR #91, opened 2016-06-29
 Remove all defaulting to localhost / port 8080

rubenvp8510:


lib/hawkular/alerts/alerts_api.rb
```diff
@@ -14,7 +14,7 @@ module Hawkular::Alerts
   # @param credentials [Hash{String=>String}] Hash of username, password, token(optional)
   # @param options [Hash{String=>String}] Additional rest client options
   class AlertsClient < Hawkular::BaseClient
-    def initialize(entrypoint = 'http://localhost:8080/hawkular/alerts', credentials = {}, options = {})
+    def initialize(entrypoint, credentials = {}, options = {})
```
pilhuhn, line 17: What happens if one does `AlertsClient.new('localhost:8080', credentials, options)` ? I think that in this case the constructor would need to internally attach the `/hawkular/alerts` part.

Similar for other sub-clients.

rubenvp8510, line 17: Ok, it makes sense, I'll do the change..

spec/integration/operations_spec.rb
```diff
@@ -119,7 +120,7 @@ module Hawkular::Operations::RSpec
     before(:all) do
       VCR.use_cassette('Operation/Helpers/get_tenant', decode_compressed_response: true) do
         @creds = { username: 'jdoe', password: 'password' }
-        inventory_client = InventoryClient.create(credentials: @creds)
+        inventory_client = InventoryClient.create(entrypoint: INVENTORY_ENTRYPOINT, credentials: @creds)
```
pilhuhn: Shouldn't that only pass in `HOST` now?

spec/unit/client_spec.rb
```diff
@@ -21,7 +17,7 @@
     end
 
     it 'should define subcomponents' do
-      client = Hawkular::Metrics::Client.new
+      client = Hawkular::Metrics::Client.new 'http://localhost:8080/hawkular/metrics'
```
pilhuhn: Same here?

lib/hawkular/base_client.rb
```diff
@@ -127,6 +127,25 @@ def generate_query_params(params = {})
       end
     end
 
+    # Generate a new url with the passed sufix path if the path is not already added
+    # to the URL
+    # @param entrypoint [String] base path
+    # @param sufix_path [String] sufix path to be added if it doesn't exist
+    # @return [String] URL with path attached to it at the end
+    def normalize_entrypoint_url(entrypoint, sufix_path)
+      relative_path_rgx = Regexp.new("#{Regexp.quote(sufix_path)}(\/)*$")
```
pilhuhn: Can you please add a dedicated unit-test for this. Especially with combinations of 
http://lcalhost:8080
http://lcalhost:8080/
http://lcalhost:8080/hawkular/alerts
http://lcalhost:8080/hawkular/alerts/

and also permutations of the suffix path like
/hawkular/alerts/
/hawkular/alerts
hawkular/alerts

abonas: I also recommend to test: IP instead of hostname, without port (when it's 80 it's typically omitted)

lib/hawkular/inventory/inventory_api.rb
```diff
@@ -25,9 +26,9 @@ def initialize(entrypoint = nil, credentials = {}, options = {})
     #   entrypoint: http://localhost:8080/hawkular/inventory
     # and another sub-hash containing the hash with username[String], password[String], token(optional)
     def self.create(hash)
-      hash[:entrypoint] ||= 'http://localhost:8080/hawkular/inventory'
       hash[:credentials] ||= {}
       hash[:options] ||= {}
+      fail 'no parameter ":entrypoint" given' if hash[:entrypoint].nil?
```
abonas: this can move 2 lines up, because if it fails, no point to do the credentials/options initialization

lib/hawkular/inventory/inventory_api.rb
```diff
@@ -25,9 +26,9 @@ def initialize(entrypoint = nil, credentials = {}, options = {})
     #   entrypoint: http://localhost:8080/hawkular/inventory
     # and another sub-hash containing the hash with username[String], password[String], token(optional)
     def self.create(hash)
-      hash[:entrypoint] ||= 'http://localhost:8080/hawkular/inventory'
       hash[:credentials] ||= {}
```
abonas, line 30: perhaps worth to unify the code which does the entrypoint/credentials assignment among subclients? it feels like copy paste... @pilhuhn

lib/hawkular/base_client.rb
```diff
@@ -127,6 +127,28 @@ def generate_query_params(params = {})
...
+    # also, this function always remove the slash at the end of the URL, so if your entrypoint is
+    # http://localhost/hawkular/inventory/ this function will return http://localhost/hawkular/inventory
+    # to the URL
+    # @param entrypoint [String] base path
+    # @param suffix_path [String] sufix path to be added if it doesn't exist
+    # @return [String] URL with path attached to it at the end
+    def normalize_entrypoint_url(entrypoint, suffix_path)
+      strip_path = suffix_path.gsub(%r{/$}, '')
+      strip_path.nil? || suffix_path = strip_path
+      strip_path = suffix_path.gsub(%r{^/}, '')
```
pilhuhn, line 140: I wonder it it would make sense to use the Uri class for this instead of doing all the (scaring :-) regex-magic.

rubenvp8510, line 140: My first approach was using URI class, but I saw that this class doesn't address well the problem of regularizing the path in the case of presence or absence of slashes at the end/begin of the path.

```
(URI.parse 'hawkular/alerts').path => "hawkular/alerts" 
(URI.parse '/hawkular/alerts').path  => "/hawkular/alerts" 
(URI.parse 'http://localhost/hawkular/alerts').path => "/hawkular/alerts" 
(URI.parse 'http://localhost/hawkular/alerts/').path => "/hawkular/alerts/" 
```

Also, I had a problem in the tests with URLs with double slashes like `http://localhost/hawkular/metrics//data` thats why I need to make sure that this method generated a URL in a uniform way.

I don't know if I missing something.

pilhuhn, line 140: This is probably more braindead, than expected, so leave it as is.

PR #97, opened 2016-07-01
Fix missing firing_match and auto_resolve_match fields on Trigger

lucasponce:
This PR fixes a couple of fields needed on Trigger to define the type of matching.
This fix is needed for ManageIQ Middleware alerting work, so once merged a new version will be needed on MiQ to continue the alerting work.

### DIFF
--- a/lib/hawkular/alerts/alerts_api.rb
+++ b/lib/hawkular/alerts/alerts_api.rb
@@ -322,7 +322,7 @@ module Hawkular::Alerts
     attr_accessor :auto_resolve, :auto_resolve_alerts, :tags, :type
     attr_accessor :tenant, :description, :group, :severity, :event_type, :event_category, :member_of, :data_id_map
     attr_reader :conditions, :dampenings
-    attr_accessor :enabled, :actions
+    attr_accessor :enabled, :actions, :firing_match, :auto_resolve_match
 
     def initialize(trigger_hash)
       return if trigger_hash.nil?
@@ -348,6 +348,8 @@ module Hawkular::Alerts
       @context = trigger_hash['context']
       @type = trigger_hash['type']
       @tags = trigger_hash['tags']
+      @firing_match = trigger_hash['firingMatch']
+      @auto_resolve_match = trigger_hash['autoResolveMatch']
       # acts = trigger_hash['actions']
       # acts.each { |a| @actions.push(Action.new(a)) } unless acts.nil?
     end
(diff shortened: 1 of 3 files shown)


lucasponce:
@pilhuhn, I have seen there is a pending PR in MiQ with hawkular-client-ruby update.
Is there a chance that we could create a release soon and use that PR to upgrade the gem with this fix ?

pilhuhn:
@lucasponce That PR is merged meanwhile. We will create a 2.2.0 gem and submit the gem update as a separate PR, which is very likely to be quickly merged.

lib/hawkular/alerts/alerts_api.rb
```diff
@@ -348,6 +348,8 @@ def initialize(trigger_hash)
       @context = trigger_hash['context']
       @type = trigger_hash['type']
       @tags = trigger_hash['tags']
+      @firing_match = trigger_hash['firingMatch']
+      @auto_resolve_match = trigger_hash['autoResolveMatch']
```
abonas, line 352: camel case is not very "Rubyish" (unlike Java), so unless these 2 parameters are coming from somewhere else that I missed, it is better to have lowercase and "_" kind of format.

lib/hawkular/alerts/alerts_api.rb
```diff
@@ -348,6 +348,8 @@ def initialize(trigger_hash)
       @context = trigger_hash['context']
       @type = trigger_hash['type']
       @tags = trigger_hash['tags']
+      @firing_match = trigger_hash['firingMatch']
```
abonas, line 351: should there be a default value for firing_match/auto_resolve_match if it's empty in trigger_hash (or a validation for illegal value) ?

lib/hawkular/alerts/alerts_api.rb
pilhuhn, line 352: Those params are parsed out of what the Hawkular server sends, which is 
camel case

Am 3. Juli 2016 3:42:28 nachm. schrieb abonas notifications@github.com:

lucasponce:
Can the 2.2.0 version being released these days ?
This patch is necessary for the pending PRs on ManageIQ and due the slow rhythm of review/merge there is a risk that alerting work will stall for another long break.
A 2.2.0 release/upgrade would help on that sense.

lib/hawkular/alerts/alerts_api.rb
lucasponce, line 351: Default values and validation are performed on server side.

PR #98, opened 2016-07-01
Use the /deprecated endpoint when Hawkular-Inventory is 0.17 or higher.

josejulio:


pilhuhn:
Close+reopen to trigger a new run against rack 1.x

pilhuhn:
Note, that this has to work:

```
  $hawkular = ::Hawkular::Client.new(hash)

  $inventory_client = $hawkular.inventory
  $inventory_client.list_feeds
```

If you need a server to test against, you can either compile hawkular-services master

josejulio:
I'm yet testing with a live server.

josejulio:
It passes the tests. I tried to run against a live server but I keep getting: 
  Hawkular-Tenant HTTP header must be provided
Tried adding the Tenant, but the tests are currently dependend of the setup (Must be tenant 28026b36-8fe4-4332-84c8-524e173a68bf, that tenant must have 1 feed, that feed 2 resources, etc.

I however tested that it returned meaningful values on 0.17.1.Final.

pilhuhn:
I verified it with the upcoming hawkular-services release that uses inventory 0.17.2 and also with an old one (~ 0.0.2 or 0.0.3) which uses inventory pre 0.17

PR #99, opened 2016-07-01
Fix rack version at 1.6.4 to support ruby <v2.2.2

pilhuhn:
Rack 2.0.1 has been released which requires Ruby >= 2.2.2

Let's continue using 1.6.4 so that also older Rubies can be supported.

### DIFF
--- a/hawkularclient.gemspec
+++ b/hawkularclient.gemspec
@@ -33,6 +33,7 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency('vcr')
   gem.add_development_dependency('rubocop', '= 0.34.2')
   gem.add_development_dependency('coveralls')
+  gem.add_development_dependency('rack', '= 1.6.4')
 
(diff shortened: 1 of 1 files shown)


abonas:
@pilhuhn 
I'm late to the party, but why specifically 1.6.4 and with '=' and not '~>' to allow support of more rack versions?
https://robots.thoughtbot.com/rubys-pessimistic-operator

pilhuhn:
1.6.4 is the version we know that it works.
We can still change to pessimistic operator to support 1.6.5.
I wanted to make sure that this is good to work with.

Am 3. Juli 2016 2:37:46 nachm. schrieb abonas notifications@github.com:

abonas:
> 1.6.4 is the version we know that it works.
> We can still change to pessimistic operator to support 1.6.5.
> I wanted to make sure that this is good to work with.

Typically I am also cautious with those kind of things, but if you know that the ruby bump was in version X, there's no reason to limit it to one minor version imo due to that.

PR #100, opened 2016-07-04
Use the pessimistic operator as suggested in #99 for the rack version

pilhuhn:
Use the pessimistic operator as suggested in #99 for the rack version.
The problematic bump was at 2.x, so minors / z-streams for 1.x should be ok.

@abonas


abonas:
LGTM. 
it is possible to be even more flexible by doing ~>1.6 (and then it will be limited by 1.6-2.0 and not from 1.6.4 to 1.7.0) but that's up to you / based on what you investigated in regards to versions compatibility.

pilhuhn:
I think I want to be on the safe side. As 2.x is out, I also do not expect major work on 1.x - perhaps smaller bugfixes.

PR #103, opened 2016-07-04
Make metrics aware of HWKMETRICS-24

### LINKED ISSUE #53: Make metrics aware of HWKMETRICS-24
HWKMETRICS-24 makes (incompatible) changes to some metrics endpoints.
The old endpoints still exist for a while, but will go away.

We may check if h-metrics has a working version endpoint and
then decide which endpoint to call according to the version.

rubenvp8510:
Fixes #53 Make metrics aware of HWKMETRICS-24

HWKMETRICS-24 makes (incompatible) changes to some metrics endpoints.
The old endpoints still exist for a while, but will go away. We may check if h-metrics has a working version endpoint and then decide which endpoint to call according to the version.

### DIFF
--- a/lib/hawkular/metrics/metric_api.rb
+++ b/lib/hawkular/metrics/metric_api.rb
@@ -34,9 +34,10 @@ module Hawkular::Metrics
       gauges.each { |g| default_timestamp g[:data] }
       counters.each { |g| default_timestamp g[:data] }
       availabilities.each { |g| default_timestamp g[:data] }
-
       data = { gauges: gauges, counters: counters, availabilities: availabilities }
-      http_post('/metrics/data', data)
+      path = '/metrics/'
+      @legacy_api ? path << 'data' : path << 'raw'
+      http_post(path, data)
     end
 
     # Base class for accessing metric definition and data of all
@@ -50,6 +51,7 @@ module Hawkular::Metrics
         @client = client
         @type = metric_type
         @resource = resource
+        @legacy_api = client.legacy_api
       end
 
       # Create new  metric definition
@@ -105,9 +107,10 @@ module Hawkular::Metrics
       #                                        {:value => 99.9, :tags => {:tagName => "myMax"}}])
       def push_data(id, data)
         data = [data] unless data.is_a?(Array)
-
+        uri = "/#{@resource}/#{ERB::Util.url_encode(id)}/"
+        @legacy_api ? uri << 'data' : uri << 'raw'
         @client.default_timestamp data
-        @client.http_post("/#{@resource}/#{ERB::Util.url_encode(id)}/data", data)
+        @client.http_post(uri, data)
       end
 
       # Retrieve metric datapoints
@@ -150,7 +153,9 @@ module Hawkular::Metrics
       def get_data_by_tags(tags, starts: nil, ends: nil, bucketDuration: nil)
         params = { tags: tags_param(tags), start: starts,
                    end: ends, bucketDuration: bucketDuration }
-        resp = @client.http_get("/#{@resource}/data/?" + encode_params(params))
+        path = "/#{@resource}/"
+        @legacy_api ? path << 'data/?' : path << 'stats/?'
+        resp = @client.http_get(path + encode_params(params))
         resp.is_a?(Array) ? resp : [] # API returns no content (empty Hash) instead of empty array
       end
 
@@ -165,8 +170,14 @@ module Hawkular::Metrics
       private
 
       def get_data_helper(id, params)
-        resp = @client.http_get("/#{@resource}/#{ERB::Util.url_encode(id)}/data/?" +
-          encode_params(params))
+        path = "/#{@resource}/#{ERB::Util.url_encode(id)}/"
+        if @legacy_api
+          path << 'data/?'
+        else
+          (params[:bucketDuration].nil? && params[:buckets].nil?) ? path << 'raw/?' : path << 'stats/?'
+        end
+        path << encode_params(params)
+        resp = @client.http_get(path)
         resp.is_a?(Array) ? resp : [] # API returns no content (empty Hash) instead of empty array
       end
     end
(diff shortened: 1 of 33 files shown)


pilhuhn:
@rubenvp8510 Please fill in a description - even if there is a reference to some jira or a ticket in the GitHub tracker in the title or commit message.

lib/hawkular/metrics/metric_api.rb
```diff
@@ -36,7 +36,22 @@ def push_data(gauges: [], counters: [], availabilities: [])
       availabilities.each { |g| default_timestamp g[:data] }
 
       data = { gauges: gauges, counters: counters, availabilities: availabilities }
-      http_post('/metrics/data', data)
+      path = 'metrics/'
+      if new_api_supported
```
pilhuhn: Below you use the elvis operator ( ?: ) - please use one for consistency - I think in this case ?: is more appropriate

lib/hawkular/metrics/metric_api.rb
```diff
@@ -36,7 +36,22 @@ def push_data(gauges: [], counters: [], availabilities: [])
...
+        path << 'raw'
+      else
+        path << 'data'
+      end
+      http_post(path, data)
+    end
+
+    def new_api_supported
```
pilhuhn: Should that be  `new_api_supported?` or even `supports_new_api?` 

And it should perhaps moved to the `private` section

lib/hawkular/metrics/metric_api.rb
```diff
@@ -152,8 +171,14 @@ def encode_params(params)
       private
 
       def get_data_helper(id, params)
-        resp = @client.http_get("/#{@resource}/#{ERB::Util.url_encode(id)}/data/?" +
-          encode_params(params))
+        path = "/#{@resource}/#{ERB::Util.url_encode(id)}/"
+        if @new_api_supported
```
pilhuhn: Elvis

lib/hawkular/metrics/metric_api.rb
```diff
@@ -197,6 +222,9 @@ def initialize(client)
       # @return [Array[Hash]] rate points
       def get_rate(id, starts: nil, ends: nil, bucket_duration: nil)
         path = "/#{@resource}/#{ERB::Util.url_encode(id)}/rate"
+        if @new_api_supported
+          bucket_duration.nil? ? path << '/raw' : path << '/stats'
+        end
         params = { start: starts, end: ends, bucketDuration: bucket_duration }
```
pilhuhn, line 225: If `bucket_duration.nil?` should it then be passed here? Not sure if it matters though.

lib/hawkular/metrics/metric_api.rb
abonas: yes, I think private section might be better. 
also "new api" is not the best terminology, as then there might be later also "new new api" etc. 
versus just "api" (=current api) and legacy api.

spec/spec_helper.rb
```diff
@@ -12,13 +12,21 @@
 require 'json'
 
 module Hawkular::Metrics::RSpec
+  MOCKER_VERSION = '0.9.0.Final'
```
abonas: why is there a need to hardcode version here?

lib/hawkular/metrics/metric_api.rb
rubenvp8510, line 225: It doesn't matter because encode_params will eliminate parameters with nil values.

lib/hawkular/metrics/metric_api.rb
rubenvp8510: Ok, I'll rename the variable and move the code to a private section.

lib/hawkular/metrics/metric_api.rb
rubenvp8510: I'm not sure if is appropriate to use elvis operator here because inside I have an elvis operator too if I use an elvis operator in this part, I'll have nested elvis operators. I don't think it will make the code readable.

abonas:
> @rubenvp8510 Please fill in a description - even if there is a reference to some jira or a ticket in the GitHub tracker in the title or commit message.

@rubenvp8510 perhaps you missed this comment :)

lib/hawkular/metrics/metric_api.rb
```diff
@@ -34,9 +34,10 @@ def push_data(gauges: [], counters: [], availabilities: [])
       gauges.each { |g| default_timestamp g[:data] }
       counters.each { |g| default_timestamp g[:data] }
       availabilities.each { |g| default_timestamp g[:data] }
-
       data = { gauges: gauges, counters: counters, availabilities: availabilities }
-      http_post('/metrics/data', data)
+      path = 'metrics/'
```
abonas: Before this PR, the call was to `http_post('/metrics/data', data)` where the path (metrics) begins with a `/`.
And after the change, it looks like path doesn't start with the slash `/`
is this possibly introducing a regression here?

lib/hawkular/metrics/metrics_client.rb
```diff
@@ -43,6 +46,9 @@ def initialize(entrypoint,
                    options = {})
       entrypoint = normalize_entrypoint_url entrypoint, 'hawkular/metrics'
       super(entrypoint, credentials, options)
+      version = fetch_version_and_status['Implementation-Version']
+      major, minor = version.scan(/\d+/).map(&:to_i)
```
abonas: any potential need in error handling? in case `fetch_version_and_status` fails and the version var is corrupted/empty?

lib/hawkular/metrics/metric_api.rb
pilhuhn: @rubenvp8510 ^^

pilhuhn:
@rubenvp8510 Can you please look at the open comments and then please rebase to resolve conflicts?

spec/spec_helper.rb
rubenvp8510: I did this to avoid re-recording the cassettes, this also allows me to switch versions in the tests in order to test this PR.

lib/hawkular/metrics/metric_api.rb
rubenvp8510: It seems like you are right, I gonna fix this.

lib/hawkular/metrics/metrics_client.rb
rubenvp8510: if fetch_version_and_status fails, it does mean that your entry point is incorrect because as far as I understand if the service is up this shouldn't happen.

So, if the version/status is corrupted we can raise an exception here.

WDYT? @pilhuhn

pilhuhn: On 22 Jul 2016, at 17:19, Ruben Vargas wrote:

> So, if the version/status is corrupted we can raise an exception here.
> 
> WDYT? @pilhuhn

Makes sense to me.

PR #104, opened 2016-07-05
Add changes + version bump to 2.2.0

pilhuhn:
List of changes in 2.2.0

@abonas please review


CHANGES.rdoc
```diff
@@ -3,6 +3,19 @@
 This document describes the relevant changes between releases of the
 _hawkular-client_ project.
 
+=== V 2.2.0
+
+This release is needed to work with hawkular-services v0.0.5+
+
+* Support the changed api endpoint (/deprecated) for Hawkular-inventory 0.17+ along the old version
+* Add two fields on Alert-triggers
```
abonas: worth to mention which fields otherwise this information is not as useful as it can be

Issue #79, opened 2016-06-25
Operations: callback not always called


pilhuhn:
Suppose I pass in an incorrect parameter hash for `invoke_generic_operation`, the callback is not called, but the method bails out with a 'fail':

```
RuntimeError: Hash property resourcePath must be specified
./lib/hawkular/operations/operations_api.rb:216:in `block in check_pre_conditions'
./lib/hawkular/operations/operations_api.rb:215:in `each'
./lib/hawkular/operations/operations_api.rb:215:in `check_pre_conditions'
./lib/hawkular/operations/operations_api.rb:105:in `invoke_generic_operation'
./spec/integration/operations_spec.rb:78:in `block (2 levels) in <module:RSpec>'
```

I think the user would only expect this if no callback is defined at all.

PR #105, opened 2016-07-06
Operations: callback not always called

rubenvp8510:
Issue #79: Operations: callback not always called

### DIFF
--- a/lib/hawkular/operations/operations_api.rb
+++ b/lib/hawkular/operations/operations_api.rb
@@ -4,6 +4,8 @@ require 'json'
 
 # Adding a method `perform` for each block so that we can write nice callbacks for this client
 class Proc
+  class PerformMethodMissing
+  end
   def perform(callable, result)
     call(Class.new do
       method_name = callable.to_sym
@@ -13,7 +15,7 @@ class Proc
       # rubocop:disable Lint/NestedMethodDefinition
       # https://github.com/bbatsov/rubocop/issues/2704
       def method_missing(_method_name, *_args, &_block)
-        false
+        PerformMethodMissing
       end
       # rubocop:enable Lint/NestedMethodDefinition
     end.new)
@@ -215,7 +217,11 @@ module Hawkular::Operations
       fail 'callback must have the perform method defined. include Hawkular::Operations' unless
           callback.nil? || callback.respond_to?('perform')
       params.each do |property|
-        fail "Hash property #{property} must be specified" if hash[property].nil?
+        next unless hash[property].nil?
+        err_callback = 'You need to specify error callback'
+        err_message = "Hash property #{property} must be specified"
+        fail(ArgumentError, err_callback) if callback.nil?
+        fail(ArgumentError, err_callback) if callback.perform(:failure, err_message).equal? Proc::PerformMethodMissing
       end
     end
 
--- a/spec/integration/operations_spec.rb
+++ b/spec/integration/operations_spec.rb
@@ -86,6 +86,82 @@ module Hawkular::Operations::RSpec
       end
     end
 
+    it 'should run into error callback because bad hash parameters' do
+      WebSocketVCR.configure do |c|
+        c.hook_uris = [HOST]
+      end
+
+      WebSocketVCR.record(example, self) do
+        client = OperationsClient.new(host: HOST,
+                                      wait_time: WebSocketVCR.live? ? 1.5 : 2,
+                                      credentials: {
+                                        username: 'jdoe',
+                                        password: 'password'
+                                      },
+                                      options: {
+                                        tenant: 'hawkular'
+                                      })
+
+        noop = { operationName: 'noop' }
+
+        client.invoke_generic_operation(noop) do |on|
+          on.success do |_data|
+            fail 'This should have failed'
+          end
+          on.failure do |_error|
+          end
+        end
+      end
+    end
+
+    it 'should bail with hash property error because no callback at all' do
+      WebSocketVCR.configure do |c|
+        c.hook_uris = [HOST]
+      end
+
+      WebSocketVCR.record(example, self) do
+        client = OperationsClient.new(host: HOST,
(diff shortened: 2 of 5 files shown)


lib/hawkular/operations/operations_api.rb
```diff
@@ -215,7 +215,10 @@ def check_pre_conditions(hash = {}, params = [], &callback)
       fail 'callback must have the perform method defined. include Hawkular::Operations' unless
           callback.nil? || callback.respond_to?('perform')
       params.each do |property|
-        fail "Hash property #{property} must be specified" if hash[property].nil?
+        if hash[property].nil?
+          error_message = "Hash property #{property} must be specified"
+          callback.nil? ? fail(error_message) : callback.perform(:failure, error_message)
```
josejulio: I guess that it would be better to raise an [ArgumentError](http://apidock.com/ruby/ArgumentError) instead of the [RuntimeError of fail](http://ruby-doc.org/core-2.0.0/Kernel.html#method-i-fail)

spec/integration/operations_spec.rb
```diff
@@ -86,6 +86,62 @@ module Hawkular::Operations::RSpec
...
+                                      })
+
+        noop = { operationName: 'noop' }
+        begin
+          client.invoke_generic_operation(noop)
+        rescue
+          puts 'We got an exception and this is good'
+        else
+          fail 'Should have failed with error:Hash property must be specified'
+        end
```
josejulio: You could use something like:
expect { client.invoke_generic_operation(noop) }.to raise_error(ArgumentError)

jkremser: +1 the "We got an exception and this is good" is polluting the output of rspec

lib/hawkular/operations/operations_api.rb
```diff
@@ -215,7 +215,10 @@ def check_pre_conditions(hash = {}, params = [], &callback)
       fail 'callback must have the perform method defined. include Hawkular::Operations' unless
           callback.nil? || callback.respond_to?('perform')
       params.each do |property|
-        fail "Hash property #{property} must be specified" if hash[property].nil?
+        if hash[property].nil?
+          error_message = "Hash property #{property} must be specified"
+          callback.nil? ? fail(ArgumentError, error_message) : callback.perform(:failure, error_message)
```
abonas: I might be missing something here, but here if callback is empty then it throws an ArgumentError. A few lines up on line 216, there is "fail .... **unless** callback.nil?" check that sounds a bit contradictory.

spec/integration/operations_spec.rb
```diff
@@ -86,6 +86,56 @@ module Hawkular::Operations::RSpec
       end
     end
 
+    it 'should run into error callback because bad hash parameters' do
```
abonas, line 89: in combination with my comment above, I recommend adding a test where callback is nil

lib/hawkular/operations/operations_api.rb
pilhuhn: While you are right @abonas the situation is even more different as the `callback.perform` works on the failure callback, but that is not checked for. 

The only situation this method should `fail` is if no error-callback is given. Otherwise the error-callback should be called with the problem (message).

rubenvp8510: Just for clarify, this should fail and throw an exception only in the following cases:

```
client.invoke_generic_operation(noop) do |on|
  on.success do |_data|
    puts 'This should have failed'
  end
end
```

`client.invoke_generic_operation(noop)
`

but shouldn't fail in this case:

```
client.invoke_generic_operation(noop) do |on|
  on.success do |_data|
    puts 'This should have failed'
  end

  on.failure do |error|
    puts error
  end
end

```

Is that correct? If this is the case, I'll do the changes.

pilhuhn: Yes. It should only fail if there is no error callback at all

Am 12. Juli 2016 5:17:49 nachm. schrieb Ruben Vargas 
notifications@github.com:

spec/integration/operations_spec.rb
```diff
@@ -86,6 +86,82 @@ module Hawkular::Operations::RSpec
...
+                                      })
+
+        noop = { operationName: 'noop' }
+        expect do
+          client.invoke_generic_operation(noop) do |on|
+            on.success do |_data|
+              fail 'This should have failed'
+            end
+          end
+        end.to raise_error(ArgumentError, 'Hash property resourcePath must be specified')
```
pilhuhn: This (and above) looks odd to me - not the failing per se, but rather the given reason text.
It should have raised the `ArgumentError` because there is no callback.

spec/integration/operations_spec.rb
```diff
@@ -86,6 +86,82 @@ module Hawkular::Operations::RSpec
...
+                                        username: 'jdoe',
+                                        password: 'password'
+                                      },
+                                      options: {
+                                        tenant: 'hawkular'
+                                      })
+
+        noop = { operationName: 'noop' }
+        expect { client.invoke_generic_operation(noop) }.to raise_error(ArgumentError,
+                                                                        'Hash property resourcePath must be specified')
```
pilhuhn: (see below)

lib/hawkular/operations/operations_api.rb
```diff
@@ -215,7 +217,10 @@ def check_pre_conditions(hash = {}, params = [], &callback)
       fail 'callback must have the perform method defined. include Hawkular::Operations' unless
           callback.nil? || callback.respond_to?('perform')
       params.each do |property|
-        fail "Hash property #{property} must be specified" if hash[property].nil?
+        next unless hash[property].nil?
+        err_message = "Hash property #{property} must be specified"
+        fail(ArgumentError, err_message) if callback.nil?
```
pilhuhn: See below - if all callbacks are missing or only the error callback is missing, we need to fail, saying so.
Once the error callback is present, all failures need to be returned via the error callback.

Issue #58, opened 2016-04-27
Catch html auth failures and return sensible values to the caller


pilhuhn:
If you run

```
    it 'Should err on bad credentials' do
      @creds = {
          username: '-XX-X-jdoe-X',
          password: 'password'
      }
      @client = Hawkular::Inventory::InventoryClient.create(credentials: @creds)

      @client.list_feeds

    end
```

Then you will see that the exception on console is some html crap from WF, that does not honor the supplied accept header.

We need to catch that and return something that makes sense.
Similar if we run in a connection exception (wrong port, host unreachable, ... )

See https://github.com/ManageIQ/manageiq/issues/8239

jkremser:
right, it can be replicated also with curl

``` bash
curl --user jdoea:password -s -H "Accept: application/json" -H "Content-Type: application/json" 'http://localhost:8080/hawkular/inventory/feeds/' | less
```

It's probably Keycloak, not the inventory itself, I can look into it. Other option would be to have some kind of interceptor in the ruby gem layer.

pilhuhn:
Let's leave this open for the moment and revisit once Hawkular-services has transitioned to use JAAS

jpkrohling:
This is not Keycloak-specific, this is returned by Wildfly itself. So, this will still be true for a JAAS scenario. Ideally, the party making the HTTP call would check the HTTP status code before trying to parse the response body.

PR #106, opened 2016-07-07
Issue #58 Catch html auth failures and return sensible values to the caller

rubenvp8510:
Issue #58 Catch html auth failures and return sensible values to the caller

### DIFF
--- a/lib/hawkular/base_client.rb
+++ b/lib/hawkular/base_client.rb
@@ -174,6 +174,8 @@ module Hawkular
     end
 
     def handle_fault(f)
+      http_code = (f.respond_to?(:http_code) ? f.http_code : 0)
+      fail HawkularException.new('Unauthorized', http_code) if f.instance_of? RestClient::Unauthorized
       if f.respond_to?(:http_body) && !f.http_body.nil?
         begin
           json_body = JSON.parse(f.http_body)
@@ -181,7 +183,7 @@ module Hawkular
         rescue JSON::ParserError
           fault_message = f.http_body
         end
-        fail HawkularException.new(fault_message, (f.respond_to?(:http_code) ? f.http_code : 0))
+        fail HawkularException.new(fault_message, http_code)
       else
         fail f
       end
--- a/spec/integration/hawkular_client_spec.rb
+++ b/spec/integration/hawkular_client_spec.rb
@@ -22,6 +22,18 @@ module Hawkular::Client::RSpec
       }
     end
 
+    it 'Should err on bad credentials' do
+      VCR.use_cassette('HawkularClient/Should err on bad credentials') do
+        @creds = {
+          username: '-XX-X-jdoe-X',
+          password: 'password'
+        }
+        expect do
+          Hawkular::Client.new(entrypoint: HOST, credentials: @creds)
+        end.to raise_error(Hawkular::BaseClient::HawkularException, 'Unauthorized')
+      end
+    end
+
     it 'Should fail when calling method with unknown prefix' do
       expect { @hawkular_client.ynventori_list_feeds }.to raise_error(RuntimeError)
       expect { @hawkular_client.list_feeds }.to raise_error(RuntimeError)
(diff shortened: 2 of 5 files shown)


lib/hawkular/base_client.rb
```diff
@@ -174,14 +174,16 @@ def tenant_header
     end
 
     def handle_fault(f)
+      http_code = (f.respond_to?(:http_code) ? f.http_code : 0)
```
abonas, line 177: when would this not have http code?
also, if http_code would be set to 0, the exception on line 186 will be confusing.

lib/hawkular/base_client.rb
```diff
@@ -174,14 +174,16 @@ def tenant_header
...
+      fail HawkularException.new('Invalid credentials', http_code) if f.instance_of? RestClient::Unauthorized
```
abonas: what about other cases when the credentials _are_ valid but don't have the right permissions? (like unauthorized suggests)

abonas:
in order to properly link & close issues when the PR is merged, please follow this guide:
https://help.github.com/articles/closing-issues-via-commit-messages/

lib/hawkular/base_client.rb
rubenvp8510, line 177: I saw this validation in the code and assume that it could happen. but it seems to me like this shouldn't happen.

lib/hawkular/base_client.rb
rubenvp8510: I can change the message here to "Unauthorized Access".

abonas: Shouldn't there be a message as part of the response that can be propagated? So there won't be a need to "guess" it or hardcode it

rubenvp8510: Yeah, it should be, but the problem is that the REST API returns HTML response for this cases, (this is the problem that this PR is trying to address), one thing I can do is try to parse as a JSON but if it fails I can return a generic "Unauthorized Access" message.

abonas: It sounds to me that in error cases the api should not return HTML from the server. @pilhuhn wdyt?

pilhuhn: @abonas this is an issue in WildFly - it will return HTML no matter what accepts header you send it. Which is why this PR exists at all. 
@rubenvp8510 parsing the body as JSON will not work. A simple regex could in this case 

401 Unauthorized:
`<html><head><title>Error</title></head><body>Unauthorized</body></html>`

500 with accept type of json:

```
{
  "errorMsg" : "RESTEASY003210: Could not find resource for full path: http://localhost:8080/hawkular/inventory/foo"
}
```

What happens is either a message is unauthorized and we get a 401+html from the appserver itself. Or it is authorized and then passed to the rest-layer (which is part of the application) and this then correctly handles the response types (and returns json).

rubenvp8510: I meant, try to parse the body as JSON but of course with an HTML will fail, in that case, I can return a fixed message or try to parse HTML as a regexp as you said, but honestly, I feel scary of parsing the HTML because it could change and is out of our control.

Anyway I can do the changes. :)

abonas: @pilhuhn understood the limitations of wildfly. @rubenvp8510 I am less familiar which errors return as json and which is html, so whatever you two find the best approach is :+1: from me.

rubenvp8510:
I only changed the message to "Unauthorized", I avoided parsing the HTML. @pilhuhn  WDYT?

pilhuhn:
Sounds good to me

Am 8. Juli 2016 3:51:11 nachm. schrieb Ruben Vargas notifications@github.com:

lib/hawkular/base_client.rb
pilhuhn, line 177: I think it can happen in cases, where the exception is e.g. a "dns-exception" before the actual http exchange happens.

PR #107, opened 2016-07-08
Fix regression where URIs as input for the entrypoint no longer worked.

pilhuhn:
Some normalization logic in 2.2.0 introduced a regression where URI:HTTP object were no longer accepted as input for entrypoints.
This pr checks for uri objects and converts them to string if needed.

@simon3z @yaacov

### DIFF
--- a/lib/hawkular/base_client.rb
+++ b/lib/hawkular/base_client.rb
@@ -140,10 +140,11 @@ module Hawkular
       strip_path.nil? || suffix_path = strip_path
       strip_path = suffix_path.gsub(%r{^/}, '')
       strip_path.nil? || suffix_path = strip_path
-      strip_entrypoint = entrypoint.to_s.gsub(%r{/$}, '')
-      strip_path.nil? && strip_entrypoint = entrypoint.to_s
+      entrypoint = entrypoint.to_s
+      strip_entrypoint = entrypoint.gsub(%r{/$}, '')
+      strip_path.nil? && strip_entrypoint = entrypoint
       relative_path_rgx = Regexp.new("\/#{Regexp.quote(suffix_path)}(\/)*$")
-      if relative_path_rgx.match(entrypoint.to_s)
+      if relative_path_rgx.match(entrypoint)
         strip_entrypoint
       else
         "#{strip_entrypoint}/#{suffix_path}"
--- a/lib/hawkular/base_client.rb
+++ b/lib/hawkular/base_client.rb
@@ -131,18 +131,19 @@ module Hawkular
     # also, this function always remove the slash at the end of the URL, so if your entrypoint is
     # http://localhost/hawkular/inventory/ this function will return http://localhost/hawkular/inventory
     # to the URL
-    # @param entrypoint [String] base path
+    # @param entrypoint [String] base path (URIs are also accepted)
     # @param suffix_path [String] sufix path to be added if it doesn't exist
     # @return [String] URL with path attached to it at the end
     def normalize_entrypoint_url(entrypoint, suffix_path)
+      fail ArgumentError, 'suffix_path must not be empty' if suffix_path.empty?
       strip_path = suffix_path.gsub(%r{/$}, '')
       strip_path.nil? || suffix_path = strip_path
       strip_path = suffix_path.gsub(%r{^/}, '')
       strip_path.nil? || suffix_path = strip_path
-      strip_entrypoint = entrypoint.gsub(%r{/$}, '')
-      strip_path.nil? && strip_entrypoint = entrypoint
+      strip_entrypoint = entrypoint.to_s.gsub(%r{/$}, '')
+      strip_path.nil? && strip_entrypoint = entrypoint.to_s
       relative_path_rgx = Regexp.new("\/#{Regexp.quote(suffix_path)}(\/)*$")
-      if relative_path_rgx.match(entrypoint)
+      if relative_path_rgx.match(entrypoint.to_s)
         strip_entrypoint
       else
         "#{strip_entrypoint}/#{suffix_path}"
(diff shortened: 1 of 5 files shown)


lib/hawkular/base_client.rb
```diff
@@ -139,10 +139,11 @@ def normalize_entrypoint_url(entrypoint, suffix_path)
       strip_path.nil? || suffix_path = strip_path
       strip_path = suffix_path.gsub(%r{^/}, '')
       strip_path.nil? || suffix_path = strip_path
-      strip_entrypoint = entrypoint.gsub(%r{/$}, '')
+      the_entry_point = (entrypoint.is_a? URI::HTTP) ? entrypoint.to_s : entrypoint
```
chessbyte: Why not the simpler `entrypoint = entrypoint.to_s`?

pilhuhn: On 8 Jul 2016, at 14:02, Oleg Barenboim wrote:

> Why not the simpler `entrypoint = entrypoint.to_s`?

Isn't a to_s on something that is already a string a bit odd? But sure I 
can change that.

chessbyte: Idiomatic Ruby /cc @Fryguy

lib/hawkular/base_client.rb
```diff
@@ -131,18 +131,18 @@ def generate_query_params(params = {})
...
     # @return [String] URL with path attached to it at the end
     def normalize_entrypoint_url(entrypoint, suffix_path)
       strip_path = suffix_path.gsub(%r{/$}, '')
       strip_path.nil? || suffix_path = strip_path
       strip_path = suffix_path.gsub(%r{^/}, '')
       strip_path.nil? || suffix_path = strip_path
-      strip_entrypoint = entrypoint.gsub(%r{/$}, '')
-      strip_path.nil? && strip_entrypoint = entrypoint
+      strip_entrypoint = entrypoint.to_s.gsub(%r{/$}, '')
+      strip_path.nil? && strip_entrypoint = entrypoint.to_s
```
abonas: what about empty strings? (I mean '')

abonas:
is this applying to all subclients or only main client and metrics?

lib/hawkular/base_client.rb
pilhuhn: The internal clients use that method to get to the correct base-endpoint, so I don't think this is an issue that could arise. I can add a guard on `suffix_path` being empty to warn the developer of a new sub-client - the `nil` case is easily caught by `nil.gsub`bailing out :-)

pilhuhn:
@abonas when you do `Hawkular::Client.new` then it applies to all sub-clients. As CM is calling it differently, I did the explicit check for their use case.

lib/hawkular/base_client.rb
```diff
@@ -131,18 +131,19 @@ def generate_query_params(params = {})
...
     # @return [String] URL with path attached to it at the end
     def normalize_entrypoint_url(entrypoint, suffix_path)
+      fail ArgumentError, 'suffix_path must not be empty' if suffix_path.empty?
       strip_path = suffix_path.gsub(%r{/$}, '')
       strip_path.nil? || suffix_path = strip_path
       strip_path = suffix_path.gsub(%r{^/}, '')
       strip_path.nil? || suffix_path = strip_path
-      strip_entrypoint = entrypoint.gsub(%r{/$}, '')
-      strip_path.nil? && strip_entrypoint = entrypoint
+      strip_entrypoint = entrypoint.to_s.gsub(%r{/$}, '')
```
chessbyte: I think doing `entrypoint = entrypoint.to_s` above here should eliminate need to do the `to_s` conversions below.

yaacov:
@pilhuhn is is better to do `client = Hawkular::Client.new ...` and then do `client.metrics...` then the `Hawkular::Metrics::Client.new ...` we use now ?
I can add it to my todo list :-)

pilhuhn:
@yaacov I don't think it matters too much in your use-case. Once you need more than one sub-client, it is better to use `Hawkular::Client.new` and getting the sub-clients from it.

@yaacov did you have a chance to test your code with this change?

lib/hawkular/base_client.rb
```diff
@@ -131,16 +131,18 @@ def generate_query_params(params = {})
...
     def normalize_entrypoint_url(entrypoint, suffix_path)
+      fail ArgumentError, 'suffix_path must not be empty' if suffix_path.empty?
       strip_path = suffix_path.gsub(%r{/$}, '')
       strip_path.nil? || suffix_path = strip_path
       strip_path = suffix_path.gsub(%r{^/}, '')
       strip_path.nil? || suffix_path = strip_path
+      entrypoint = entrypoint.to_s
       strip_entrypoint = entrypoint.gsub(%r{/$}, '')
-      strip_path.nil? && strip_entrypoint = entrypoint
+      strip_path.nil? && strip_entrypoint = entrypoint.to_s
```
yaacov: in line 143 `entrypoint = entrypoint.to_s` is converted to string, do we need the `.to_s` here ?

I checked in CM and this works fine :-) with and without this `.to_s`

pilhuhn: you need the `to_s` as the entrypoint could be a `URI::HTTP` (what you had originally) and without it the `gsub()` in line 144 will error out as you have reported :)

yaacov: > entrypoint could be a URI::HTTP

after line 143 entrypoint is string

pilhuhn: Now I get what you mean - sorry

josejulio: ~~There is a `to_s` in line 143 and other in line 145. I guess the later isn't needed, right?~~

yaacov: np. I checked it without the `to_s` in line 145 and CM metrics wroks :-)

pilhuhn:
Ok, when Travis is happy, I will merge it and then release the 2.2.1 version of the gem and PR it on MiQ side

PR #111, opened 2016-07-11
Metric ID not escaped in MetricClient.update_tags

### LINKED ISSUE #109: Metric ID not escaped in MetricClient.update_tags
The metric:

```
get "http://localhost:8080/hawkular/metrics/gauges/MI~R~%5b70c798a0-6985-4f8a-a525-012d8d28e8a3%2fLocal~~%5d~MT~WildFly%20Aggregated%20Web%20Metrics~Aggregated%20Active%20Web%20Sessions", "Accept"=>"application/json", "Accept-Encoding"=>"gzip, deflate", "Content-Type"=>"application/json", "Hawkular-tenant"=>"hawkular"
# => 200 OK | application/json 238 bytes
{"id":"MI~R~[70c798a0-6985-4f8a-a525-012d8d28e8a3/Local~~]~MT~WildFly Aggregated Web Metrics~Aggregated Active Web Sessions"
```

Exception when calling `MetricClient.update_tags`
`bad URI(is not URI?): http://localhost:8080/hawkular/metrics/gauges/MI~R~[70c798a0-6985-4f8a-a525-012d8d28e8a3/Local~~]~MT~WildFly Aggregated Web Metrics~Aggregated Active Web Sessions/tags
`
( see spaces and [ and / characters )

also potentially applies to other methods.

rubenvp8510:
Sometimes the metric ID contains special characters like`"/ & ?"` in some methods, those characters had not escaped properly.  When RestClient tries to parse the URL, it fails with an exception `bad URI(is not URI?):`

Fixes #109

### DIFF
--- a/lib/hawkular/metrics/metric_api.rb
+++ b/lib/hawkular/metrics/metric_api.rb
@@ -79,14 +79,15 @@ module Hawkular::Metrics
       # @param id [String]
       # @return [MetricDefinition]
       def get(id)
-        the_id = @client.hawk_escape id
+        the_id = ERB::Util.url_encode id
         Hawkular::Metrics::MetricDefinition.new(@client.http_get("/#{@resource}/#{the_id}"))
       end
 
       # update tags for given metric definition
       # @param metric_definition [MetricDefinition]
       def update_tags(metric_definition)
-        @client.http_put("/#{@resource}/#{metric_definition.id}/tags", metric_definition.hash[:tags])
+        metric_definition_id = ERB::Util.url_encode metric_definition.id
+        @client.http_put("/#{@resource}/#{metric_definition_id}/tags", metric_definition.hash[:tags])
       end
 
       # Push metric data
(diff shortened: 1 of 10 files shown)


lib/hawkular/metrics/metric_api.rb
```diff
@@ -79,14 +79,15 @@ def query(tags = nil)
       # @param id [String]
       # @return [MetricDefinition]
       def get(id)
-        the_id = @client.hawk_escape id
```
pilhuhn, line 82: While I have no objection in general to replace `hawk_escape` by `url_encode` I think it should be done in one go and consistently.

pilhuhn:
Besides the above, LGTM

pilhuhn:
@abonas WDYT?

lib/hawkular/metrics/metric_api.rb
abonas, line 82: an alternative way to write this would be adding in the head of the file
`require "erb"`
`include ERB::Util`

and then just call in the code `url_encode` 
it does exactly the same thing, but just is shorter

abonas:
left an optimization comment above, besides that LGTM. great work @rubenvp8510

PR #113, opened 2016-07-14
Inventory restapi 0.17.0.final

jkremser:
This is originally Lukas' work taken from his [branch](https://github.com/metlos/hawkular-client-ruby/tree/inventory-0.17.0.Final). I had to fix couple of minor issues and make the test suite passing.

Now the client supports both the inventory < 0.17 - the old REST api and inventory > 0.17 (the new API, not the deprecated endpoint).

Some tests now have weaker checks to satisfy both contexts. There are also some ugly switches caused mostly by incompatibilities in the JSON payloads.

Now, it's possible to call the rspec w/ env property describing the version of the inventory, we are testing, by default the latest (currently 0.17) is used (micro version is ignored).

Example: `INVENTORY_VERSION=0.16.2.Final rspec`   ... run the tests for the old api

^ this was actually added as another Rake task so that it can be also run by `rake old-inventory`, travis will run also this task


lib/hawkular/inventory/inventory_api.rb
```diff
@@ -20,21 +22,9 @@ def initialize(entrypoint = nil, credentials = {}, options = {})
...
-        begin
-          backup_entrypoint = @entrypoint
-          @entrypoint = @entrypoint[0...-'/deprecated'.size]
-          return http_get('/status')
-        ensure
-          @entrypoint = backup_entrypoint
-        end
-      end
-      http_get('/status')
+      @version = version.scan(/\d+/).map(&:to_i)
```
pilhuhn, line 25: Can this be collapsed as in previous line 23? I don't see `@version` to be used further.
Looks like you may be using that in the spec?

abonas:
1. is this proposing any risks/backwards incompatibility if brought into miq?
2. please update the gem readme as part of the PR with relevant supported versions
   thanks

spec/integration/inventory_spec.rb
```diff
@@ -279,8 +319,11 @@ module Hawkular::Inventory::RSpec
       resource_path = CanonicalPath.new(feed_id: feed_id, resource_ids: [hawk_escape_id('Local~~')])
       config = @client.get_config_data_for_resource(resource_path)
 
+      # product name is missing
       expect(config['value']['Server State']).to eq('running')
-      expect(config['value']['Product Name']).to eq('Hawkular')
```
pilhuhn, line 283: This and similar is a bit unfortunate, as the change is not specific to inventory, but rather to the old recording against a version that identified as Hawkular vs what we have currently.

Similar for 'ExampleDS' vs 'Datasource[ExampleDS]'  If we would re-record against a current server with the old api, we would see ''Datasource[ExampleDS]'  in both cases.

pilhuhn:
This is a huge, needed change. 
The new api is in since hawkular-services 0.0.5 (a week ago). People should use this anyway and the old hawkular-full is no longer supported.

I think I would feel more comfortable to not pollute the spec by version switches (and probably not even the api) and just move to the new api, keeping spec changes as small as possible.
This way we can more easily track the changes and validate them.

I need to do some other stuff, but can then also pull that PR branch and play with it in HawkFX and MiQ to see if I detect fallout.

lib/hawkular/inventory/inventory_api.rb
jkremser, line 25: It's being used, for instance [here](https://github.com/hawkular/hawkular-client-ruby/pull/113/files#diff-2f01ee789318dc3149af508f5d2947d7R20), [here](https://github.com/hawkular/hawkular-client-ruby/pull/113/files#diff-2f01ee789318dc3149af508f5d2947d7R98), etc.

spec/integration/inventory_spec.rb
jkremser, line 283: that's correct. The question is what makes the API, to some extend it's also the data that are being sent, or at least their format. I agree, the tests w/ those 'ifs' look cumbersome

> a current server with the old api

That's pretty artificial, isn't it? Current server is current, because it has the current versions of component with their current APIs :] If you mean current APIs with old data, that would require changing the agent version, which won't work w/o additional effort.

jkremser:
> **not pollute the spec by version switches**

Ok, I'll do my best, I don't like it either. Now, I am fighting with the Coveralls to report the coverage data for all the rake tasks, not just the last one

jkremser:
> 1. is this proposing any risks/backwards incompatibility if brought into miq?

I'd like to say 100% no, but it needs to be tested also in MiQ. I hit some incompatibilities in the data, but the current approach in the gem that uses the `/deprecated` endpoint will see the same data.

jkremser:
Closing this PR, because it was decided that the old Inventory API will not be supported. There will be a new PR soon.

PR #114, opened 2016-07-14
Inventory restapi 0.17.2.final

jkremser:
v 2.0

I've smoke tested the MiQ with this merged in and it works.


lib/hawkular/inventory/entities.rb
```diff
@@ -0,0 +1,217 @@
+# It contains class definitions that are used by the inventory REST client
+module Hawkular::Inventory
+  # A Basic inventory entity with id, name, path and optional properties
+  class BaseEntity
+    # @return [String] Full path of the entity
+    attr_reader :path
+    # @return [String] Name of the entity
+    attr_reader :name
+    # @return [String] Name of the entity
```
abonas: this should be id and not Name in the comment (a copy paste from line 7 :smile) but is there really a need for comments for things like name/id? what is a `@` sign in the comment for?

lib/hawkular/inventory/entities.rb
```diff
@@ -0,0 +1,217 @@
...
+    attr_reader :id
+    # @return [String] Feed this entity belongs to (or nil in case of a feedless entity)
+    attr_reader :feed
+    # @return [String] Name of the environment for this entity
```
abonas, line 13: what are the potential values of env property?

lib/hawkular/inventory/entities.rb
```diff
@@ -0,0 +1,217 @@
...
+    attr_reader :env
+    # @return [String] Properties of this entity
+    attr_reader :properties
+
+    def initialize(hash)
+      @id = hash['id']
+      @path = hash['path']
+      @name = hash['name'] || @id
+      @properties = hash['properties'] || {}
+      @_hash = hash.dup
```
abonas, line 23: underscore is not very Rubyish :smile: 
also, why is there a need to dup the hash?

lib/hawkular/inventory/entities.rb
```diff
@@ -0,0 +1,217 @@
...
+        case key
+        when 'f'
+          @feed = val
+        when 'e'
+          @env = val
+        end
+      end
+    end
+
+    def ==(other)
```
abonas, line 39: Please elaborate what is this method for. Also, inside it uses `eql?` which is pretty similar in most cases to `==`. perhaps you meant to use `equal?` (reference equality check) ? (and if there is a need to override `==`, please check - perhaps `eql?` needs to be overriden too)

lib/hawkular/inventory/entities.rb
```diff
@@ -0,0 +1,217 @@
...
+    def initialize(type_hash)
+      super(type_hash)
+      @type = type_hash['type']
+      @unit = type_hash['unit']
+      @collection_interval = type_hash['collectionInterval']
+    end
+  end
+
+  # Definition of a Metric inside the inventory.
+  class Metric < BaseEntity
```
abonas, line 93: it seems that Metric and MetricType are very much alike - is there a way to reuse the code rather than copy paste it?

lib/hawkular/inventory/entities.rb
jkremser, line 13: It takes any values, but by default, it's `test` (more here https://git.io/vKrzP). I mean the `test` environment is pre-created in inventory for each feed (/agent).

lib/hawkular/inventory/entities.rb
jkremser, line 23: Right, it's there from the Libor's times. To be honest, I don't know. What it does is a deep copy of the hash that is passed to the constructor. This looks like a good idea, because if you do this

``` ruby
def foo(hash)
  hash['foo'] = 0
  hash
end
my_hash = { 'foo' => 1 }
foo(my_hash)
puts my_hash .. will print.. { 'foo' => 0 }
```

I guess, the `@_hash` is there to make the `to_h` method faster (+less code to write)

lib/hawkular/inventory/entities.rb
jkremser, line 39: Right, it should be using `equal?` instead of `eql?` ("hash idenity"), in this case it's the same, but in general you are right.

Why is it there? I guess, the same reason as in Java.. to define where 2 objects are equal, I think in spec we sometimes fetch the entity with properties and sometimes not, and it shouldn't matter, because what makes 2 object equal is the `id` in this case

Oh, and it's not my code, it was there before. All this PR did (regarding this piece of code) is splitting it into a new file to separate the client logic from the entity classes definition.

lib/hawkular/inventory/entities.rb
jkremser, line 93: Hmm, it's tempting to do `Metric < MetricType` and in the Metric constructor call `super(metric_hash['type'])`.

But do we want to have this `Metric  --- is_a ----> MetricType` relation? It may be dangerous, what if api consumers calls those tests? Introducing a new (private) class and have both Metric and MetricType extend it. Hmm, imho doesn't worth doing it, but I can do that if you insist.

Btw. same holds for `Resource` and `ResourceType` and again, this wasn't part of this PR.

lib/hawkular/inventory/entities.rb
pilhuhn, line 23: Yes. To all of that.

pilhuhn:
I am merging this, so we can go forward and test it in local miq installs and/or HawkFX.
We can make the "cosmetic" changes in followup PRs.

pilhuhn:
Thanks @metlos and @Jiri-Kremser

lib/hawkular/inventory/entities.rb
abonas, line 39: OK so you can fix it here or in a separate PR if this is not a new code. your call.
And `equal?` vs `eql?` is definitely confusing especially coming from Java...

lib/hawkular/inventory/entities.rb
abonas, line 93: I don't insist on anything :smiley: 
I'm just trying to see if copy paste can be avoided.
perhaps a mixin would be suitable here?

PR #115, opened 2016-07-15
Wraps connection errors on HawkularConnectionException.

josejulio:


lib/hawkular/base_client.rb
```diff
@@ -175,6 +178,16 @@ def tenant_header
       headers
     end
 
+    # @!visibility private
```
abonas, line 181: could you explain what is this comment/attribute? 
(btw, based on line 169 this is already part of a private section)

lib/hawkular/base_client.rb
```diff
@@ -175,6 +178,16 @@ def tenant_header
...
+    def connect_error(fault)
+      if fault.is_a?(SocketError)
+        HawkularConnectionException.new(fault.to_s)
+      elsif [Errno::ECONNREFUSED, Errno::ECONNRESET, Errno::EHOSTUNREACH,
```
abonas, line 185: my concern here is the need to manually maintain a list of those error types

lib/hawkular/base_client.rb
```diff
@@ -175,6 +178,16 @@ def tenant_header
...
+             Errno::ENETDOWN, Errno::ENETUNREACH, Errno::ETIMEDOUT].include?(fault.class)
+        HawkularConnectionException.new(fault.to_s, fault.class::Errno)
```
abonas, line 187: this is very similar to line 184. Could you please elaborate what do we gain here by adding the fault.class vs the previous assignment in the "if" part where it is not added to the exception?

lib/hawkular/base_client.rb
```diff
@@ -163,6 +163,9 @@ def initialize(message, status_code = 0)
       attr_reader :message, :status_code
     end
 
+    class HawkularConnectionException < HawkularException
```
abonas, line 166: What is the benefit in adding a custom exception for the gem? it indicates a standard connection error that is not unique for hawkular (or to be more accurate, multiple different situations of connection issues)

lib/hawkular/base_client.rb
josejulio, line 181: This is something that I copy/pasted, I guess that doesn't do anything. I will remove if not needed.

lib/hawkular/base_client.rb
josejulio, line 187: The socket error doesn't seem to have any status code. It happened when trying to connect to an unknown hostname.
The others had the error code as a class constant and I'm adding them as part of the Exception.

lib/hawkular/base_client.rb
josejulio, line 185: I was looking for a way to not have to, but IIRC the parent class of all of these is StandardError. It seems that needs to be handled this way.

lib/hawkular/base_client.rb
pilhuhn, line 181: See http://www.rubydoc.info/gems/yard/file/docs/GettingStarted.md ( also for the @ in `@return` of another PR.
I can't tell if this is appropriate for this specific line. Probably not if there is not more documentation of the method.

lib/hawkular/base_client.rb
abonas, line 187: I think you can just check if the status code is not nil and then add it. but the current code doesn't add the status code if I am not mistaken?

lib/hawkular/base_client.rb
abonas, line 181: imo there's no much point to have it here but I am not strongly opinionated regarding this

lib/hawkular/base_client.rb
abonas, line 187: another thing that concerns me is that the docs say the names of the errors vary per operation system
http://ruby-doc.org/core-2.2.0/Errno.html
so I'd like to propose to reconsider again - do we need such specific handling for connection errors? this definitely doesn't sound unique to hawkular. 
cc @pilhuhn

pilhuhn:
I thought I discussed that last week, but don't find my recordings :)  I believe, that it does not really matter what exception is thrown here but rather that we catch it in the MiQ code while we are looping over the servers.

lib/hawkular/base_client.rb
josejulio, line 166: There seems to be the possible exceptions that are throw on unix / windows

http://ruby-doc.org/stdlib-1.9.2/libdoc/socket/rdoc/Socket.html#method-i-connect

josejulio:
How should we proceed on this one? 

I found out that [handling](http://tammersaleh.com/posts/rescuing-net-http-exceptions/) the exceptions from Net:Http is actually similar to what is proposed here, [rest-client](https://github.com/rest-client/rest-client/blob/master/lib/restclient/request.rb#L479) (the gem being used by hawkular-client-ruby) uses Net:Http. There are [alternatives](http://stackoverflow.com/a/5370726/688379) to this, but would mean to change rest-client or catch the [connection exceptions](http://ruby-doc.org/stdlib-1.9.2/libdoc/socket/rdoc/Socket.html#method-i-connect) from Net:Http.

lib/hawkular/base_client.rb
pilhuhn, line 187: I think we are safe for all that are POSIX compliant ( http://man7.org/linux/man-pages/man3/errno.3.html ). Windows since NT is Posix as well as real Unixoids (like Linux).

abonas, line 187: so it sounds like this PR is ready for merge, @pilhuhn ?
(and then after releasing a new gem version @josejulio can upgrade the gem in miq and use it in #9835) . wdyt?

PR #118, opened 2016-07-22
Ensures entrypoint is a String before applying a substitution.

### LINKED ISSUE #116: Another gsub does not work on URL issue
See https://github.com/hawkular/hawkular-client-ruby/blob/master/lib/hawkular/hawkular_client.rb#L61

```
  Operations::OperationsClient.new(entrypoint: @state[:entrypoint].gsub(/^https?/, 'ws'),
```

josejulio:
Fixes #116

### DIFF
--- a/spec/integration/hawkular_client_spec.rb
+++ b/spec/integration/hawkular_client_spec.rb
@@ -267,6 +267,15 @@ module Hawkular::Client::RSpec
         end
       end
 
+      it 'Should work initializing with URI' do
+        uri = URI.parse HOST
+        opts = { tenant: 'hawkular' }
+        WebSocketVCR.record(example, self) do
+          the_client = Hawkular::Client.new(entrypoint: uri, credentials: @creds, options: opts)
+          expect { the_client.operations }.to_not raise_error
+        end
+      end
+
       xit 'Should both reuse the websocket connection', :websocket do
         WebSocketVCR.record(example, self) do
         end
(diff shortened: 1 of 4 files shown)


spec/integration/hawkular_client_spec.rb
```diff
@@ -267,6 +267,18 @@ module Hawkular::Client::RSpec
...
       end
 
+      it 'Should work initializing with URI' do
+        uri = URI.parse HOST
+        opts = { tenant: 'hawkular' }
+        WebSocketVCR.configure do |c|
+          c.hook_uris = [HOST]
+        end
+        WebSocketVCR.record(example, self) do
+          the_client = Hawkular::Client.new(entrypoint: uri, credentials: @creds, options: opts)
```
pilhuhn, line 274: I think this is the critical part of the Issue/PR.

spec/integration/hawkular_client_spec.rb
```diff
@@ -267,6 +267,18 @@ module Hawkular::Client::RSpec
...
+          expect { the_client.operations }.to_not raise_error
```
pilhuhn, line 275: Here you need a running server to test/do the initial recording. The cassette provided is empty. 
Perhaps @Jiri-Kremser can help you if needed.

jkremser, line 275: The cassette is empty, because the `WebSocketVCR` records only the websocket communication, but in this case the WS connection wasn't opened. It's opened lazily during the first use of `the_client.operations` (https://git.io/vKMEh)

So either call the `the_client.operations` to open the WS connection, or remove the `WebSocketVCR.record(example, self) do`, this way it doesn't make much sense.

josejulio:
I recorded the cassette with the wrong "hook_uri" on the WebSocketVCR. 
Fixed it with help of @Jiri-Kremser

pilhuhn:
Thank you both.

PR #119, opened 2016-07-22
Drying the Metric and MetricType

jkremser:
Placing the common fields to a module and mix them in to the Metric and MetricType classes to make the code DRY.

cc @abonas

### DIFF
--- a/lib/hawkular/inventory/entities.rb
+++ b/lib/hawkular/inventory/entities.rb
@@ -73,13 +73,21 @@ module Hawkular::Inventory
     end
   end
 
-  class MetricType < BaseEntity
+  # Fields that are common for MetricType and Metric
+  module MetricFields
     # @return [String] GAUGE, COUNTER, etc.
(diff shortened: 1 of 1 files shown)


lib/hawkular/inventory/entities.rb
```diff
@@ -73,13 +73,21 @@ def initialize(res_hash)
...
 
-  class MetricType < BaseEntity
+  # Fields that are common for MetricType and Metric
+  module MetricFields
     # @return [String] GAUGE, COUNTER, etc.
     attr_reader :type
     # @return [String] metric unit such as NONE, BYTES, etc.
     attr_reader :unit
-    # @return [Long] collection interval in seconds
+    # @return [Long] collection interval in seconds, it has different semantics for MetricType and for Metric
```
abonas, line 82: nit - missing a "." char, otherwise this and next lines are read together. I had to reread it to understand without punctuation.

PR #121, opened 2016-07-26
Make it explicit that we require rest client v2.

pilhuhn:
This follows ManageIQ.

@abonas Can you please verify and eventually merge?


abonas:
why do we have to limit it in the gem? bringing the gem to miq will limit it anyway, no?

pilhuhn:
On 26 Jul 2016, at 13:23, abonas wrote:

> why do we have to limit it in the gem? bringing the gem to miq will 
> limit it anyway, no?

In the past we did not really know if it would work or not. Making
it explicit allows to answer that more easily. But again, I don't mind.

abonas:
verified.

PR #122, opened 2016-07-26
Allow to pull in developer-local gems via Gemfile.dev.rb.

pilhuhn:
@abonas Can you please verify and eventually merge?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 /tmp/
 spec/vcr.log
 Gemfile.lock
+Gemfile.dev.rb
 
 ## Specific to RubyMotion:
 .dat*
--- a/Gemfile
+++ b/Gemfile
@@ -1,2 +1,4 @@
 source 'http://rubygems.org'
 gemspec
+dev_gemfile = File.expand_path('Gemfile.dev.rb', __dir__)
+eval_gemfile(dev_gemfile) if File.exist?(dev_gemfile)
--- a/README.rdoc
+++ b/README.rdoc
@@ -148,3 +148,7 @@ Client documentation can be generated using http://yardoc.org
 
     yardoc
 
+== Developer-local gems
+
+It is now possible to have a local gemfile `Gemfile.dev.rb` that is pulled in from the
+main `Gemfile` to use e.g. debuggers or rubocop-git.


abonas:
I'm wondering if this is the right thing to do in a gem (vs a rails app). Gemfile is typically not used in gem development afaik.

pilhuhn:
On 26 Jul 2016, at 11:55, abonas wrote:

> I'm wondering if this is the right thing to do in a gem (vs a rails 
> app). Gemfile is typically not used in gem development afaik.

That Gemfile.dev.rb is certainly not checked in and relying on it
would fail on travis.

I have a commit-hook that runs rubocop-git and without that
developer-gemfile it fails. Of course we could add it to the main
gemspec, but it feels like polluting it.

Dunno

## 

hwr@pilhuhn.de  m:0179/207 4919  b:http://pilhuhn.blogspot.com

abonas:
I totally understand why it's needed, but just unsure that this is the right way to do it in a gem.
@jrafanie @durandom - wdyt?

.gitignore
```diff
@@ -10,6 +10,7 @@
 /tmp/
 spec/vcr.log
 Gemfile.lock
+Gemfile.dev.rb
```
jrafanie, line 13: 👍

jrafanie:
> I'm wondering if this is the right thing to do in a gem (vs a rails app). Gemfile is typically not used in gem development afaik.

@abonas @pilhuhn I'm familiar with other gems that do this.  Basically, you put a Gemfile in the root and call gemspec in it and add other things you need as a developer of the gem.  The Gemfile.dev.rb can be for things you don't want to force other developers of the gem to use or "bundle".  Gemfile.dev.rb can be for debuggers, profilers, and other things that aren't easily installed on all platforms.

[rest-client](https://github.com/rest-client/rest-client/blob/3a79728b332d316afa281d10022658735a6f72df/Gemfile)

[psych](https://github.com/tenderlove/psych/blob/e10e35f3e0ef91926dd7ac4b1192e25f24d0f741/Gemfile)

[fog](https://github.com/fog/fog/blob/e1a9c6237ece04a587241baa4ce4f7ae9a8d400a/Gemfile)

abonas:
verified with and without the dev file.

PR #123, opened 2016-07-26
Release notes and version bump for 2.3.0.

pilhuhn:
@abonas can you please check and eventually merge?

### DIFF
--- a/CHANGES.rdoc
+++ b/CHANGES.rdoc
@@ -3,6 +3,18 @@
 This document describes the relevant changes between releases of the
 _hawkular-client_ project.
 
+=== V 2.3.0
+
+* Wrap connection errors in `HawkularConnectionException` (a subclass of `HawkularException`)
+* Make sure the error callback is called on all kinds of errors (if it exists)
+* Use the new Rest-API of Hawkular-Inventory.
+* For authentication errors don't return raw html, but some sensible value
+* Make it explicit that rest-client v2.0 is used to follow ManageIQ
+* Provide parameters for operation definitions (depend on what the agent reports to inventory)
+* Use the new Hawkular-metrics endpoints if the hawkular-metrics version we talk to is >= 0.16
+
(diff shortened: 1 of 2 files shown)


lib/hawkular/version.rb
```diff
@@ -4,5 +4,5 @@
 # @see https://github.com/hawkular
 module Hawkular
   # Version of the Hawkular Ruby Gem
-  VERSION = '2.2.1'
+  VERSION = '2.3.0'.freeze
```
abonas, line 7: why does it need a freeze?

CHANGES.rdoc
```diff
@@ -3,6 +3,18 @@
 This document describes the relevant changes between releases of the
 _hawkular-client_ project.
 
+=== V 2.3.0
+
+* Wrap connection errors in `HawkularConnectionException` (a subclass of `HawkularException`)
+* Make sure the error callback is called on all kinds of errors (if it exists)
+* Use the new Rest-API of Hawkular-Inventory.
+* For authentication errors don't return raw html, but some sensible value
+* Make it explicit that rest-client v2.0 is used to follow ManageIQ
```
abonas, line 12: this is not merged yet (I left a comment on that PR...)
btw, looking forward we can try using [this gem](https://github.com/skywinder/Github-Changelog-Generator) to get the list of changes automatically. wdyt?

lib/hawkular/version.rb
pilhuhn, line 7: On 26 Jul 2016, at 13:24, abonas wrote:

> > @@ -4,5 +4,5 @@
> >  # @see https://github.com/hawkular
> >  module Hawkular
> >    # Version of the Hawkular Ruby Gem
> > -  VERSION = '2.2.1'
> > -  VERSION = '2.3.0'.freeze
> 
> why does it need a freeze?

I think Rubocop suggested it. Isn't it good practice
to freeze the "constants"?
But then idc

abonas, line 7: never saw a freeze on a version - it shouldn't matter in this case imo, but I guess it's harmless

PR #124, opened 2016-07-26
Expose metric.type.id

josejulio:
Before, the metric.name was also the same as the metric.type.id, but since [this](http://lists.jboss.org/pipermail/hawkular-dev/2016-July/002973.html) (see first point) it become a human readable value.

In ManageIQ we used said name to create a [mapping](https://github.com/ManageIQ/manageiq/blob/master/product/live_metrics/middleware_server.yaml). As we no longer can trust that value it needs to be updated to use the metric.type.id. Thus we need to expose it.

### DIFF
--- a/hawkularclient.gemspec
+++ b/hawkularclient.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency('websocket-client-simple', '~> 0.3.0')
   gem.add_runtime_dependency('addressable')
   gem.add_development_dependency('shoulda')
-  gem.add_development_dependency('rspec-rails', '~> 3.0')
+  gem.add_development_dependency('rspec-rails', '~> 3.1')
   gem.add_development_dependency('rake', '< 11')
   gem.add_development_dependency('simple-websocket-vcr', '= 0.0.7')
   gem.add_development_dependency('yard')
--- a/lib/hawkular/inventory/entities.rb
+++ b/lib/hawkular/inventory/entities.rb
@@ -101,10 +101,13 @@ module Hawkular::Inventory
   class Metric < BaseEntity
     include MetricFields
 
+    attr_reader :type_id
+
     def initialize(metric_hash)
       super(metric_hash)
       @type = metric_hash['type']['type']
       @type_path = metric_hash['type']['path']
+      @type_id = metric_hash['type']['id']
       @unit = metric_hash['type']['unit']
       @collection_interval = metric_hash['type']['collectionInterval']
     end
--- a/spec/integration/inventory_spec.rb
+++ b/spec/integration/inventory_spec.rb
@@ -292,6 +292,15 @@ module Hawkular::Inventory::RSpec
       expect(metrics.size).to be(14)
     end
 
+    it 'Should have the same requested metric type id' do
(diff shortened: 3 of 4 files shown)


pilhuhn:
Can we - possibly with help from @jmazzite add tests that document
what should be returned when.
Especially to see how it should behave if this new metric.id is
present or absent?

josejulio:
Sure, I'll add tests on it. 
I'm not sure how long has the metric.type.id been there. I guess that since a lot of time, is just that we didn't used it.

josejulio:
I added a simple test on it. Not sure if anything else is needed.

On the current release of hawkular-agent, the metric.name == metric.type.id; but it will change on future releases.

pilhuhn:
The Ruby 2.2 build is failing. We need to figure out what is wrong there - especially as MiQ is currently on Ruby 2.2

abonas:
> The Ruby 2.2 build is failing. We need to figure out what is wrong there - especially as MiQ is currently on Ruby 2.2

this is really weird. I'm trying to figure it out.

josejulio:
@rubenvp8510  pointed to me this out: http://stackoverflow.com/questions/31106024/rspec-have-attributes-matcher

I was checking and ruby2.2 has: 3.0.4 and the others 3.5.1.

@pilhuhn Is possible to raise that version? I could also change the test...

josejulio:
I meant to say travis ruby2.2; I tested it locally (with both ruby 2.2.0 and 2.2.5) and it passes.

abonas:
@rubenvp8510 @josejulio very nice catch. I am looking now at the gemspec and I am wondering why rspec-rails is listed there and not plain rspec. @Jiri-Kremser perhaps you have any idea?

jkremser:
> ..wondering why rspec-rails is listed there and not plain rspec

No idea, it's there since the very first commit

josejulio:
Tried bumping the rspec-rails to 3.5 but it can't find a dependency on ruby2.0 an 2.1. So I'm using 3.1 it seems to work.

abonas:
@pilhuhn can you please review? it LGTM from technical pov, but I am less familiar with metrics part to address functionality.
I'll see if rspec-rails can be switched to plain rspec (separately from this PR) - don't see why a plain ruby gem should have this dependency, but need to tests for failures.

pilhuhn:
I am not convinced this is enough. The metric-type is like the class definition and the metric is the instantiated object.

@jmazzitelli wrote in that email you cite:

> Metric Type definitions created by the agent will have the same ID as before, but its Name is changing
> [..]
> Clients have to know to look for the "metric-id" property and use it when looking up metric data in h-metrics 

So to be able to graph the metrics we need to have both
- type-id to find the mapping from the yaml definition to do the mapping
- pull that 'metric-id' property and use this to do the fetching of the metric.

And as the 'metric-id' can be "random" we can't just deduct it from type-id or type-name, but need to explicitly expose that to users/clients.

Unless I misunderstand something.

josejulio:
You are right, but I'm not addressing the metric-id on this PR, only exposing the id of the metric type. 

I could extend it to address it, but I didn't because there is still a pending discussion wether the agent should update the inventory-metric-id to match the metric-id.

PR #125, opened 2016-07-27
Revert "Allow to pull in developer-local gems via Gemfile.dev.rb."

abonas:
Reverts hawkular/hawkular-client-ruby#122

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,6 @@
 /tmp/
 spec/vcr.log
 Gemfile.lock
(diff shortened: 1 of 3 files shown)


abonas:
cc @pilhuhn the build got broken when merging the relevant PR for adding gemfile.dev.rb. although the error seemed unrelated to the change, I reverted it to be on the safe side.
the error can be seen here:
https://travis-ci.org/hawkular/hawkular-client-ruby/jobs/147730499

PR #126, opened 2016-08-04
hwkagent-120 Deployment undeploy required UndeployApplicationRequest

jshaughn:
Provide the infrastructure to perform deployment removal via the dedicated UndeployApplicationRequest/Response messages and not as a generic operation.  Note that this is done via the server level and not directly on the deployment itself.
- Add CanonicalPath.up utility method

@pilhuhn please review

### DIFF
--- a/lib/hawkular/inventory/entities.rb
+++ b/lib/hawkular/inventory/entities.rb
@@ -178,37 +178,37 @@ module Hawkular::Inventory
       @metric_id = hash[:metric_id]
     end
 
-    # rubocop:disable Metrics/CyclomaticComplexity
     def self.parse(path)
       fail 'CanonicalPath must not be nil or empty' if path.to_s.strip.length == 0
-      tmp = path.split('/')
-      hash = {}
-      tmp.each do |pair|
-        (key, val) = pair.split(';')
-        case key
-        when 't'
-          hash[:tenant_id] = val
-        when 'f'
-          hash[:feed_id] = val
-        when 'e'
-          hash[:environment_id] = val
-        when 'm'
-          hash[:metric_id] = val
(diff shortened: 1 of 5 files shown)


lib/hawkular/inventory/entities.rb
```diff
@@ -178,39 +178,37 @@ def initialize(hash)
...
-          hash[:resource_ids].push(val)
-        when 'mt'
-          hash[:metric_type_id] = val
-        when 'rt'
-          hash[:resource_type_id] = val
-        end
+      CanonicalPath.new(path_to_h path)
+    end
+
+    def up
```
pilhuhn, line 188: What does `up` stand for?

jshaughn, line 188: It's like CanonicalPath.up() in Inventory's Java class: " Returns a path corresponding to the direct ancestor of the resource represented by this path object."  I was just adding the same API bit to this ruby class.

pilhuhn, line 188: On 5 Aug 2016, at 15:29, Jay Shaughnessy wrote:

> > +
> > -    def up
> 
> It's like CanonicalPath.up() in Inventory's Java class: " Returns a 
> path corresponding to the direct ancestor of the resource represented 
> by this path object."

Thanks. Could we name that 'parent_path' and/or add some description?

jshaughn, line 188: Sure, I'll add some doc.  I'll keep the method name the same to have parity with the java api.  I expect we can build up this ruby impl in the same way, on an as-needed basis. Commit coming...

pilhuhn:
Thank you @jshaughn

PR #128, opened 2016-08-09
Updates Inventory and operations to work on secure connections

josejulio:
Updated inventory specs to test on a secure connection.
Inventory specs now delete the feeds created for testing.


josejulio:
Should I disable secure connection checking by default or expect that everyone setups the secure connection when creating more specs?

josejulio:
@pilhuhn @abonas  could you please share your thoughts on this?

josejulio:
Should I open a PR for each ? (metrics, inventory, alerts, operations)

pilhuhn:
Perhaps one PR for metrics and one for the rest - the metrics one allows the CM team to (only) review that one, as they use the metrics sub-client.

Disabling and enabling by default has both issues. If/when Hawkular-server can easily be set up to support ssl and/or supports SSL by default, then SSL checks should be on. Perhaps we need to just re-visit this in the future.

josejulio:
> Perhaps one PR for metrics and one for the rest - the metrics one allows the CM team to (only) review that one, as they use the metrics sub-client.

Ok

> Disabling and enabling by default has both issues. If/when Hawkular-server can easily be set up to support ssl and/or supports SSL by default, then SSL checks should be on. Perhaps we need to just re-visit this in the future.

I'll have it disabled by default in the meantime.

lib/hawkular/inventory/inventory_api.rb
```diff
@@ -485,7 +485,8 @@ def events(type = 'resource', action = 'created')
       }
       ws_options[:headers][:'Hawkular-Tenant'] = tenant_id
 
-      url = "#{entrypoint.gsub(/https?/, 'ws')}/ws/events?tenantId=#{tenant_id}&type=#{type}&action=#{action}"
+      url = "#{entrypoint.sub(/https/, 'wss').sub(/http/, 'ws')}/ws/events?"\
+      "tenantId=#{tenant_id}&type=#{type}&action=#{action}"
```
pilhuhn: Please indent here. Or split the entrypoint schema manipulation in its own statement

spec/integration/inventory_spec.rb
```diff
@@ -3,569 +3,611 @@
...
-          username: '-XX-X-jdoe-X',
-          password: 'password'
-        }
-        expect do
-          Hawkular::Inventory::InventoryClient.create(entrypoint: ENTRYPOINT, credentials: @creds)
-        end.to raise_error(Hawkular::BaseClient::HawkularException, 'Unauthorized')
-      end
-    end
-  end
+  [NON_SECURE_CONTEXT, SECURE_CONTEXT].each do |secure_context|
```
pilhuhn: Souldn't the variable rather be `context` (or `used_context`) and not `secure_context`, as it can be both secure and non-secure?

pilhuhn:
@Jiri-Kremser Can you have an eye on this please as you are the one who initially did all the VCR version magic?

spec/integration/inventory_spec.rb
jkremser: `context` is a keyword from rspec. `security_context` could work

spec/integration/inventory_spec.rb
```diff
@@ -3,569 +3,611 @@
...
+          expect(types.size).to be >= 18
+        end
 
-      expect(resources.size).to be(1)
-      resource = resources[0]
-      expect(resource.instance_of? Hawkular::Inventory::Resource).to be_truthy
-      # depends how pinger is fast
-      expect(2..6).to cover(resource.properties.size)
-      expect(resource.properties['url']).to eq(URL_RESOURCE)
-    end
```
jkremser, line 217: This should now work after https://github.com/hawkular/hawkular-services/pull/51 was merged. But I am ok with omitting it from the specs, because we don't use it anymore. @pilhuhn wdyt? do we want to still test that after url is correctly added and possibly if the pinger pings it? Is a spec in ruby gem the right place to do that?

pilhuhn, line 217: On 10 Aug 2016, at 12:47, Jirka Kremser wrote:

> But I am ok with omitting it from the specs, because we don't use it 
> anymore. @pilhuhn wdyt? do we want to still test that after url is 
> correctly added and possibly if the pinger pings it? Is a spec in ruby 
> gem the right place to do that?

Pinger is dead. So I don't think we need this url-handling special
code at all anymore.

spec/integration/inventory_spec.rb
```diff
@@ -3,569 +3,611 @@
...
+  end
 
-    it 'Should return the version' do
-      data = @client.fetch_version_and_status
-      expect(data).not_to be_nil
+  describe 'Inventory' do
+    it 'Should list feeds when using SSL without certificate disabling the verify' do
+      # change this to the real credentials when updating the VCR
+      super_secret_username = 'username'
+      super_secret_password = 'password'
```
jkremser: Would it be possible to take those from the config? Like in here https://github.com/hawkular/hawkular-client-ruby/blob/master/spec/spec_helper.rb#L18:L19 ? I am not sure what is better, I guess the user needs to change those anyway. Btw. why `jdoe:password` that is pre-created in the dev build doesn't work here?

jkremser:
Good job! General question, does this PR assumes that the CA cert is imported in the system? Travis is green, so it doesn't run it, right?

spec/integration/inventory_spec.rb
josejulio: torii was down by the time i was recording, so left it as it and didn't record a new cassette for that. 

I could take those values from the config as I'm going to modify all the specs i could try to organize the endpoint.yml.

josejulio:
> Good job!

Thank you!

> General question, does this PR assumes that the CA cert is imported in the system?

Yes, it does assume that the CA is already on the default ca_path.

On fedora I did this by copying the cert to /etc/ssl/certs and calling c_rehash (and for it to work with Hawkfx I had to call c_rehash -old -n because [jruby uses an old openssl](https://github.com/jruby/jruby/issues/2010#issuecomment-57146383))

> Travis is green, so it doesn't run it, right?

 I think that since the cassette is already recorded the VCR doesn't even open a request to query the endpoint. thus no cert checking at all on tests with VCR.

lib/hawkular/inventory/inventory_api.rb
josejulio: [Made a separate method](https://github.com/hawkular/hawkular-client-ruby/pull/129) to handle the scheme manipulation, since I'm sure i'll use it elsewhere and I don't want to be using sub directly.

spec/integration/inventory_spec.rb
josejulio: Agreed, will use security_context

spec/integration/inventory_spec.rb
josejulio: > Would it be possible to take those from the config? Like in here https://github.com/hawkular/hawkular-client-ruby/blob/master/spec/spec_helper.rb#L18:L19 ? I am not sure what is better, I guess the user needs to change those anyway.

Done, I'm using setup_inventory_client now and using record instead of VCR.use_cassette.

>  Btw. why jdoe:password that is pre-created in the dev build doesn't work here?
> Updated the cassette to use jdoe/password

josejulio:
Thanks for the feedback, i updated the code to reflect the comments.
This PR depents on #129 

edit:
Just to clarify, the :SecureConnection tests don't run anymore by default, until is ssl comes out-of-the-box or is easily setup.

jkremser:
> I think that since the cassette is already recorded the VCR doesn't even open a request..

right

spec/spec_helper.rb
```diff
@@ -195,22 +199,60 @@ def record(prefix, bindings, explicit_cassette_name, example: nil)
...
+    }
+    options[:erb] = bindings if bindings
+    if ENV['VCR_UPDATE'] == '1'
+      options[:record] = :all
+      options[:reverse_substitution] = true if bindings
+    end
+    WebSocketVCR.use_cassette(prefix + '/' + explicit_cassette_name, options) do
+      run_lambda.call
+    end
+  end
```
josejulio, line 266: @Jiri-Kremser can you take a look at this record_websocket and record_websocket_cassette methods ? Thanks

jkremser, line 266: lgtm, perhaps [this](https://github.com/josejulio/hawkular-client-ruby/blob/ecda95aa1e6f48eba99ee5c48e7c101b0a621566/spec/integration/inventory_spec.rb#L493) is worth rewriting with your helper function, but it shouldn't block this PR

josejulio:
@Jiri-Kremser @rubenvp8510 @abonas 
Please review

PR #130, opened 2016-08-12
Updated metrics specs to test on both secure and non secure connections.

josejulio:
Disabled secure connections by default, use SKIP_SECURE_CONTEXT=0 to run these.
Disabled secure connections on metrics_0.8.0.

@pilhuhn Do you think I should install a metrics0.8.0 and run secure connection tests on these or is ok to skip?


josejulio:
@pilhuhn could you take a look?

PR #133, opened 2016-08-17
Don't try to parse params if no properties are passed. Fixes #131

### LINKED ISSUE #131: List operations sometimes fails
@jmazzitelli and I saw this fly by:

```
Exception in thread "JavaFX Application Thread" org.jruby.exceptions.RaiseException: (NoMethodError) undefined method `[]' for nil:NilClass
    at RUBY.initialize(/Users/hrupp/.rvm/gems/jruby-9.1.2.0/gems/hawkular-client-2.4.0/lib/hawkular/inventory/entities.rb:122)
    at RUBY.block in list_operation_definitions(/Users/hrupp/.rvm/gems/jruby-9.1.2.0/gems/hawkular-client-2.4.0/lib/hawkular/inventory/inventory_api.rb:418)
    at org.jruby.RubyArray.each(org/jruby/RubyArray.java:1593)
    at RUBY.list_operation_definitions(/Users/hrupp/.rvm/gems/jruby-9.1.2.0/gems/hawkular-client-2.4.0/lib/hawkular/inventory/inventory_api.rb:417)
    at RUBY.list_operation_definitions_for_resource(/Users/hrupp/.rvm/gems/jruby-9.1.2.0/gems/hawkular-client-2.4.0/lib/hawkular/inventory/inventory_api.rb:429)
    at RUBY.add_operations(/Users/hrupp/src/hawkfx/lib/on_click_cell_factory.rb:130)
    at RUBY.block in initialize(/Users/hrupp/src/hawkfx/lib/on_click_cell_factory.rb:51)
```

Looking at the code shows `param_list = op_hash['properties']['params']`,
so it looks like the received entity sometimes has no properties (or worse: H-inventory sends us some empty hashes).

pilhuhn:
Newer versions of inventory may not send properties hash if there are none.
Before an empty hash was sent.

cc @jmazzitelli

### DIFF
--- a/lib/hawkular/inventory/entities.rb
+++ b/lib/hawkular/inventory/entities.rb
@@ -119,6 +119,7 @@ module Hawkular::Inventory
     def initialize(op_hash)
       super(op_hash)
       @params = {}
+      return if op_hash['properties'].nil?
       param_list = op_hash['properties']['params']
       return if param_list.nil?
       param_list.each do |p|
--- a/spec/integration/inventory_spec.rb
+++ b/spec/integration/inventory_spec.rb
@@ -333,6 +333,8 @@ module Hawkular::Inventory::RSpec
       expect(shutdown_def.params).to include('restart')
       restart_param = shutdown_def.params.fetch 'restart'
       expect(restart_param['type']).to eq('bool')
+      resume_def = operation_definitions.fetch 'Resume'
+      expect(resume_def.params).to be {}
     end
 
     it 'Should list operation definitions of given resource' do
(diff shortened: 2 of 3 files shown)


lib/hawkular/inventory/entities.rb
```diff
@@ -119,6 +119,7 @@ class OperationDefinition < BaseEntity
     def initialize(op_hash)
       super(op_hash)
       @params = {}
+      return if op_hash['properties'].nil?
```
abonas, line 122: maybe worth to check also `.empty?` ?  in case it's not nil but is an empty hash for some reason.

pilhuhn, line 122: On 17 Aug 2016, at 16:13, abonas wrote:

> maybe worth to check also `.empty?` ?  in case it's not nil but is an 
> empty hash for some reason.

If they are empty then the next check should catch
this.

abonas, line 122: ok :)

PR #136, opened 2016-08-23
Add support for fetching stats for multiple metrics

lucasponce:


lucasponce:
Hi @Jiri-Kremser, I am working on this one to add a new endpoint into the metrics client.
I have prepared a test for this but I am having problems (or perhaps I don't follow correctly the instructions) when I try to record the VCR.
How to add that ?
Running the whole tests against a live instance on localhost brings a lot of issues, running only the metrics ones also have some issues.
What is the strategy for recording new tests ?
Can you give me some advice about how add a VCR for this one (re-recording all metrics, or just one).
Could you take a quick look if errors I'm having locally are due my env or if those are well known ?
I'm using a locally hawkular-services with master of time of this PR.
Thanks !

josejulio:
You could record the test alone:
VCR_UPDATE=1 bundle exec rspec spec/integration/metric_spec.rb:198
(198 being the line on which the test start)

lucasponce:
The fear I have is if I record the test alone we could have a mixed set of different versions on the VCR.
And I am afraid that the VCR passes but when we have a live version it could fails (as I believe it happens now).
So, perhaps it is time to think about improving this process in some way.
I will send an email with general ideas about this but asking Jiri for some guidance to follow same strategy.

josejulio:
You are right about that, on #130  I'm updated the tests to also work on a secure conection, but also tried to make them easier to run at once (as i had to test all of them).

lucasponce:
In any case, I have recorded a single test for this one.
We can discuss the general topic of synchronization of VCRs in the main list.
One idea is that we could try to automate that on a new hawkular-services version, the CI could run the tests against a live instance and show if there are errors to force a re-recording.

lib/hawkular/metrics/metric_api.rb
```diff
@@ -40,6 +40,28 @@ def push_data(gauges: [], counters: [], availabilities: [])
...
+    # @param ends [Integer] optional timestamp (default now)
+    # @param bucket_duration [String] optional interval (default 3600s)
+    # @return [Hash] stats grouped per type
+    # @example
+    #   client = Hawkular::Metrics::client::new
+    #   metrics = { gauge: ['G1', 'G2'], counter: ['C2', 'C3'], availability: ['A2', 'A3'] }
+    #   starts = 200
+    #   ends = 500
+    #   bucketDuration = '150ms'
+    #   client.query_stats()
```
josejulio: The example seems to be a bit confusing, query_stats expect gauge_ids, counter_ids, avail_ids but you start to group them on a metrics hash. Could you make it a bit clearer?

lucasponce: Thanks. Forgot to update the example.

josejulio:
LGTM

jkremser:
I was trying to describe the setup at readme here https://github.com/hawkular/hawkular-client-ruby/tree/master/spec

wrt. version it's a bit tricky. The goal is that the version is dynamically obtained from the server (`/status` endpoint) and then the cassette is recorded/replayed from the right file path (the version is part of that path, hence different versions = different cassettes). Currently the first part is not there (finding out the version automatically), but the second one should be there at least for the inventory specs. So currently the version needs to be specified by `INVENTORY_VERSION` env. property. I hope I am not mistaken, I haven't checked for quite some time.

josejulio:
Thanks @lucasponce !

lucasponce:
Thanks for taking care of this @josejulio.
@abonas is there any possibility to release a gem with this to have it on manageiq ?
I see there are some other tasks on the client too.
I would need this endpoint for some enhancements I am working on livemetrics side.
Thanks.

Issue #134, opened 2016-08-18
get operation parameters via the official inventory way


jmazzitelli:
Inventory has an "official" way of storing operation parameters within operation types - a child of the operation type which is a data entity whose structural data is named "parameterTypes"

The Ruby client needs to get operation parameters using this mechanism, rather than the ad-hoc way we were doing it (via general properties).

The WildFly agent is creating both (the "official" way and the general props way) for backward compatibility, but storing these in general props is deprecated and will go away, see JIRA [HWKAGENT-130](https://issues.jboss.org/browse/HWKAGENT-130)

The new way will have inventory show you the params like this:

URL (notice "/d;parameterTypes"): http://127.0.0.1:8080/hawkular/inventory/entity/f;feed_id/rt;WildFly%20Server/ot;Shutdown/d;parameterTypes

```
{
  "path": "/t;hawkular/f;feed_id/rt;WildFly%20Server/ot;Shutdown/d;parameterTypes",
  "name": "parameterTypes",
  "identityHash": "1ad5d5963d6978975e958de5abcb49f778695f",
  "contentHash": "5ff8951bc6a6ab6483f69ec0ab139d5a76e22c",
  "syncHash": "27163827304a2c50e31b9ce2ee74e5e32f396d58",
  "value": {
    "timeout": {
      "type": "int",
      "description": "Timeout in seconds to allow active connections to drain",
      "defaultValue": "0",
      "required": false
    },
    "restart": {
      "type": "bool",
      "description": "Should the server be restarted after shutdown?",
      "defaultValue": "false",
      "required": false
    }
  }
}
```

PR #137, opened 2016-08-24
hkwagent-130 The agent now stores op params in the recommended way

jshaughn:
update handling

### DIFF
--- a/lib/hawkular/inventory/inventory_api.rb
+++ b/lib/hawkular/inventory/inventory_api.rb
@@ -412,9 +412,12 @@ module Hawkular::Inventory
       parsed_path = CanonicalPath.parse(resource_type_path.to_s)
       feed_id = parsed_path.feed_id
       resource_type_id = parsed_path.resource_type_id
-      ret = http_get("/traversal/f;#{feed_id}/rt;#{resource_type_id}/type=ot")
+      ots = http_get("/traversal/f;#{feed_id}/rt;#{resource_type_id}/type=ot")
       res = {}
-      ret.each do |ot|
+      ots.each do |ot|
+        ot_name = ERB::Util.url_encode ot['name']
+        pts = http_get("/traversal/f;#{feed_id}/rt;#{resource_type_id}/ot;#{ot_name}/d;parameterTypes")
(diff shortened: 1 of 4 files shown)


jmazzitelli:
the code to retrieve the param types looks correct to me.

lib/hawkular/inventory/inventory_api.rb
```diff
@@ -412,9 +412,12 @@ def list_operation_definitions(resource_type_path)
       parsed_path = CanonicalPath.parse(resource_type_path.to_s)
       feed_id = parsed_path.feed_id
       resource_type_id = parsed_path.resource_type_id
-      ret = http_get("/traversal/f;#{feed_id}/rt;#{resource_type_id}/type=ot")
+      ots = http_get("/traversal/f;#{feed_id}/rt;#{resource_type_id}/type=ot")
       res = {}
-      ret.each do |ot|
+      ots.each do |ot|
+        ot_name = ERB::Util.url_encode ot['name']
```
josejulio, line 418: I'm wondering why we aren't using ERB::Util.url_encode instead of hawk_escape_id and hawk_escape ?
do you know @Jiri-Kremser  ?
// cc: @rubenvp8510

rubenvp8510: IIRC these methods do the same thing (encode URL special characters).
 IMHO is better using the `ERB::Util.url_encode method`instead of the other two because is part of a well-tested ruby module.

jshaughn:
Fixes #134

lib/hawkular/inventory/inventory_api.rb
jkremser: Agreed that `ERB::Util.url_encode` is better than something hand made. The question is whether it behaves the same for all the characters as our methods.

jshaughn: In this usage it's basically there to handle spaces in an operation name, so I think it should be fine.

jshaughn:
@jmazzitelli I think this is safe to merge.  The only affected client afaik is hawkFX, and not even the master branch.

PR #140, opened 2016-09-01
Exposes metric-id property if it exists.

### LINKED ISSUE #110: when mapping inventory-metric ID to hawkular-metric ID, check if metrid-id property exists
There is a use-case to give users some freedom and flexibility to choose what their hawkular-metric IDs look like, rather than force them to adopt the rule "inventory metric ID is always the same as hawkular metric ID". 

The hawkular wildfly agent is introducing an optional "metric-id" property found on the inventory metric instance definition. When this property exists, the hawkular-metric ID is the value of that property (it is not the value of that inventory metric instance ID).

The current behavior will remain the same and will become the default - inventory metric instance ID is the same as the hawkular-metric ID - they will only be different if someone adds that "metric-id" property in inventory.

See the HWKAGENT JIRA issue and associated PR where the agent creates this metric-id when appropriate: https://issues.jboss.org/browse/HWKAGENT-78

josejulio:
Fixes #110 

I'm not sure if this is the right way to expose this id. Or if we should call it hawkular_metric_id or native_metric_id.

Other ways could to be replace the 'id' value, is the id used somewhere else?

### DIFF
--- a/lib/hawkular/inventory/entities.rb
+++ b/lib/hawkular/inventory/entities.rb
@@ -102,6 +102,8 @@ module Hawkular::Inventory
     include MetricFields
 
     attr_reader :type_id
+    # @return [String] metric id used in Hawkular Metrics
+    attr_reader :hawkular_metric_id
 
     def initialize(metric_hash)
       super(metric_hash)
@@ -110,6 +112,7 @@ module Hawkular::Inventory
       @type_id = metric_hash['type']['id']
       @unit = metric_hash['type']['unit']
       @collection_interval = metric_hash['type']['collectionInterval']
+      @hawkular_metric_id = @properties.key?('metric-id') ? @properties['metric-id'] : @id
     end
   end
 


josejulio:
@pilhuhn  could you please take a look?

pilhuhn:
LGTM. @jmazzitelli any preference on the name?

jmazzitelli:
As for the name, I don't have a preference, though "hawkular-metric-id" does seem to make sense since this does represent the "Hawkular Metrics ID". If we do change that (which seems like a good idea), we need to change hawkular-agent and hawkfx, but those changes are simple. 

On the hawkular-agent side - just change this constant string to the new value we want:

https://github.com/hawkular/hawkular-agent/blob/master/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/inventory/MeasurementInstance.java#L34

@pilhuhn we'd also have to change hawkfx - but that too should be simple - these two places:

https://github.com/pilhuhn/hawkfx/blob/master/lib/on_click_cell_factory.rb#L103

https://github.com/pilhuhn/hawkfx/blob/master/lib/chart_view_controller.rb#L53-L54

josejulio:
I already have a branch with HawkFX changes :)

jmazzitelli:
@josejulio  create a PR in hawkular-agent, and then we can press the "merge" button on all three PRs at the same time to keep everything in sync.

josejulio:
Done.
Hawkular-agent: https://github.com/hawkular/hawkular-agent/pull/252
HawkFX: https://github.com/pilhuhn/hawkfx/pull/24

PR #143, opened 2016-09-06
Add changelog for 2.7.0

lucasponce:
@josejulio @theute @rubenvp8510 
/cc @abonas 
I have added a changelog for 2.6.1 and bumped the versioning.
I have never release the gem, so I am not sure if this is enough or if we need additional steps.

Also, I have doubt if we should bump to 2.6.1 or 2.7.
Perhaps the SSL and the fix are relatively minor changes from end user.
But anyway, if criteria is to bump 2.7 please let me know and I will modify the number.

### DIFF
--- a/CHANGES.rdoc
+++ b/CHANGES.rdoc
@@ -3,6 +3,11 @@
 This document describes the relevant changes between releases of the
 _hawkular-client_ project.
 
+=== V 2.7.0
+
+* Support for SSL connections
+* Fix fetching stats to support rates
+
 === V 2.6.0
 
 * Fix fetching operation params in the recommended way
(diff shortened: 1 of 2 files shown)


josejulio:
I think that 2.7.0 is the way to go, since we added new functionality (fetching stats to support rates) and is backwards compatible.

CHANGES.rdoc
```diff
@@ -3,6 +3,11 @@
 This document describes the relevant changes between releases of the
 _hawkular-client_ project.
 
+=== V 2.7.0
+
+* Support for SSL connections
+* Fix fetching stats to support rates
+
```
abonas, line 10: missing the milestone part like below (Full list of items can be found at https://github.com/hawkular/hawkular-client-ruby/milestone/11?closed=1) but it's not critical (and not a gem requirement)

josejulio: I can assist creating the 2.7.0 milestone

https://github.com/hawkular/hawkular-client-ruby/milestone/12?closed=1

theute, line 10: Alissa did the release:
https://rubygems.org/gems/hawkular-client/versions/2.7.0
Thanks!

PR #145, opened 2016-09-20
Adding tests on a live server (via docker).

josejulio:
TODO:
- [ ] Make alerts specs aware of VCR_UPDATE

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,20 @@
+sudo: required
+
 language: ruby
 rvm:
   - 2.0
   - 2.1
   - 2.2
   - 2.3.0
+env:
+  - RUN_ON_LIVE_SERVER=1 \
+    VCR_UPDATE=1 \
+    SKIP_SSL_WITHOUT_CERTIFICATE_TEST=1 \
+    SKIP_SECURE_CONTEXT=1 \
+    SKIP_V8_METRICS=1 \
+    SKIP_SERVICES_METRICS=1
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -1,6 +1,16 @@
+sudo: required
+
 language: ruby
 rvm:
   - 2.0
```
pilhuhn, line 5: I wonder (and this is nothing for this PR specifically) if we should drop explicit support for 2.0 (and perhaps even 2.1)

.travis.yml
```diff
@@ -1,6 +1,16 @@
...
   - 2.1
   - 2.2
   - 2.3.0
+env:
+  - RUN_ON_LIVE_SERVER=1 VCR_UPDATE=1 SKIP_SSL_WITHOUT_CERTIFICATE_TEST=1 \
```
pilhuhn: Only cosmetic (and no requirement to change): I would find one per line more easy to quickly see what is set.

josejulio: You are right on that, I'll update it.

.travis.yml
josejulio: Not sure, if those versions are no longer supported/used on MIQ it would be good to trim them out.

pilhuhn: Well, that client is not only for MiQ. MiQ required 2.2.2+

pilhuhn:
Thank you @josejulio

PR #147, opened 2016-09-22
Updated alert specs

josejulio:
Working with VCR_UPDATE and easier to test on a live server.


josejulio:
@lucasponce 
I updated the alert specs, could you please take a look/review ?

// cc: @pilhuhn

spec/integration/alert-resources/events-test-data.json
```diff
@@ -0,0 +1,78 @@
+{
+  "events": [
+    {
+      "id": "my-event-01",
+      "category": "general-category",
+      "text": "some text for this event",
+      "extras": {
```
lucasponce, line 7: extras ?
in Alerting model there is a "context" field for additional data

spec/integration/alert-resources/events-test-data.json
```diff
@@ -0,0 +1,78 @@
...
+        "ctime": 2000
```
lucasponce, line 8: By the way, "ctime" is also present in the Event model of alerting

spec/integration/alerts_spec.rb
```diff
@@ -475,11 +557,38 @@ module Hawkular::Alerts::RSpec
...
-      }
+  describe 'Alert/Events' do
+    before(:all) do
+      # Setup for testing
+      record('Alert/Events', credentials, 'setup') do
+        @client = Hawkular::Alerts::AlertsClient.new(ALERTS_BASE, credentials, options)
+        json = IO.read('spec/integration/alert-resources/events-test-data.json')
+        hash = JSON.parse(json)
+        hash['events'].each do |event|
+          @client.create_event event['id'], event['category'], event['text'], event['extras']
```
lucasponce, line 568: I see here is the mapping between "extras" and what create_event uses.
It is just a naming issue but I guess is good unify terminology to not change semantic.
First think I thought reading the test json file was that there were an intermediate class with an "extras" field mapped onto an Event on Hawkular Alerts.

spec/integration/alerts_spec.rb
```diff
@@ -488,20 +597,16 @@ module Hawkular::Alerts::RSpec
       events = client.list_events('thin' => true)
 
       expect(events).to_not be_nil
-      expect(events.size).to be(12)
+      expect(events.size).to be >= 12
```
lucasponce, line 600: If events are recorded it could be good to have here an equal.

lucasponce:
@josejulio LGTM I put a couple of minor comments just in case you would like to review them.
No blockers.

spec/integration/alerts_spec.rb
josejulio: I did this so it could run on a live server, making it more tolerant to do something like VCR_UPDATE=1 bundle exec specs

spec/integration/alerts_spec.rb
josejulio: I'm using extras from because of this: https://github.com/hawkular/hawkular-client-ruby/blob/master/lib/hawkular/alerts/alerts_api.rb#L311
But i guess that for unifying the terminology it would be good to have those `extras` as named params. So the implementer wouldn't have to guess (or check the REST docs) to know what `extras` are available.

lucasponce: Yeah. I don't remember that, I'm +1 to unify terminology with the REST API then it's easier to browse through the clients. Perhaps this "extras" parameter can be better rename.
Anyway, just a side comment, non a blocker and it can be done in a future iteration IMO.

josejulio:
@lucasponce Thanks!

Could someone merge this one? @pilhuhn @rubenvp8510 
Thanks!

PR #148, opened 2016-09-22
Make some specs to wait for data before actually testing and failing.

josejulio:
Some tests might fail because travis isn't fast enough, try to wait longer in those cases.


rubenvp8510:
this LGTM.

PR #153, opened 2016-10-03
Adding yet more time to travis build

josejulio:


josejulio:
More time seems to resolve the issue. Wonder if travis will demand more time in the future...

Issue #155, opened 2016-10-06
Tenant handing in h-metrics 0.20 has changed


pilhuhn:
"rspec ./spec/integration/metric_spec.rb:100 # metrics_0_16_0 NonSecure Tenants Should create and return tenant" failure

```
"The first endpoint to be protected by this mechanism is /tenants since its functionality is cross tenant"
```

cc @yaacov - this will be a change in h-metrics for ose 3.4

PR #156, opened 2016-10-07
Fixes Issue 154

jshaughn:
- maintain API by implementing no-arg ackBy and resolvedBy methods
- re-record alerting VCRs due to JSON change, Alerts now supply Lifecycle array

@pilhuhn or @josejulio , please review


lib/hawkular/alerts/alerts_api.rb
```diff
@@ -523,8 +523,24 @@ def to_h
...
+      statusBy('ACKNOWLEDGED')
+    end
+
+    def resolvedBy
+      statusBy('RESOLVED')
+    end
+
+    def statusBy(status)
+      a = @lifecycle.nil? ? [] : @lifecycle.select { |l| l['status'].eql? status }
+      a.empty? ? '' : a.last['user']
```
josejulio: I guess it would be better to return `nil` instead of empty string.
That would help to know when an Alert hasn't been acknowledged.
`alert.resolvedBy.nil?`

spec/vcr_cassettes/Alert/Triggers/Templates/setup.yml
```diff
@@ -37,7 +37,7 @@ http_interactions:
       Pragma:
       - no-cache
       Date:
-      - Tue, 27 Sep 2016 18:17:14 GMT
+      - Fri, 07 Oct 2016 15:36:32 GMT
```
josejulio: I think that we should only update related cassettes to avoid cluttering the commit history.
WDYT @pilhuhn  ?

josejulio (changes requested):

josejulio:
Rubocop is complaining about snake_case vs camel case.
Please check output of `bundle exec rubocop`

lib/hawkular/alerts/alerts_api.rb
pilhuhn: @josejulio The goal here is to keep the api on the ruby side as-is and not changing tests. I don't know if this is the case here or not - just a general remark.

spec/vcr_cassettes/Alert/Triggers/Templates/setup.yml
pilhuhn: Yes, don't re-record if they are not changed. But I fear that is harder done than said. But such "empty" recordings could be removed before committing

josejulio: To help to only record/run specific cassettes/tests you can append the line number of the `it` to the spec command.
e.g. `VCR_UPDATE=1 bundle exec rspec spec/integration/alerts_spec:527` 
and that will only run (and re-record) that test.

lib/hawkular/alerts/alerts_api.rb
```diff
@@ -523,9 +523,29 @@ def to_h
...
+    end
+
+    def status_by(status)
+      a = @lifecycle.nil? ? [] : @lifecycle.select { |l| l['status'].eql? status }
+      a.empty? ? '' : a.last['user']
     end
+
+    # for some API back compatibility
+    alias_method :ackBy, :ack_by
+    alias_method :resolvedBy, :resolved_by
```
josejulio, line 548: thank you!

lib/hawkular/alerts/alerts_api.rb
josejulio: After a detailed inspection, It seems that it was returning null before this change. 
See ackBy on: https://github.com/hawkular/hawkular-client-ruby/pull/156/files#diff-82e612a2fe3edccd5616886ec158d0cdL105

jshaughn: that's a good question, let me verify that '' is correct.  If not I'll change to nil.

spec/vcr_cassettes/Alert/Triggers/Templates/setup.yml
jshaughn: sorry, got too aggressive, I have re-committed with only relevant re-recording.

lib/hawkular/alerts/alerts_api.rb
jshaughn: +1 JoseJulio, changing to return nil if the status has never been set.

josejulio (approved):

josejulio:
Test failures on live server are because #155

josejulio:
Thanks @jshaughn !

PR #158, opened 2016-10-10
Add an avail get_data test.

jshaughn:
It just tests an extra code path that we didn't have.

@pilhuhn please review.

### DIFF
--- a/spec/integration/metric_spec.rb
+++ b/spec/integration/metric_spec.rb
@@ -138,9 +138,10 @@ security_contexts.each do |security_context|
           id1 = SecureRandom.uuid
           id2 = SecureRandom.uuid
           id3 = SecureRandom.uuid
+          id4 = SecureRandom.uuid
 
           ids = [id1, id2, id3]
-          bindings = { id1: id1, id2: id2, id3: id3 }
+          bindings = { id1: id1, id2: id2, id3: id3, id4: id4 }
           example = proc do
             @client = setup_client(setup_options.merge tenant: vcr_test_tenant)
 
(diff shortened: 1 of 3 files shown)


josejulio:
Rubocop strikes again!
@jshaughn could you please check?

josejulio:
Travis is complaining about the cassette **metrics_services**/Templates/Mixed_metrics/Should_requests_raw_data_for_multiple_metrics.yml

I'm not really sure if we should keep metrics_0_0_16 and metrics_services. 
It seems that we are using them both alike.
// cc: @pilhuhn @rubenvp8510

rubenvp8510:
I think at this point we can remove metrics_0_0_16 and leave only metric_services.

This was for the testing  of transition from old versions of metrics to metrics 0.0.16 (mostly endpoint changes). but because now metric_services test against  recent versions of metrics, this doesn't make sense anymore.

pilhuhn:
I think it is ok to treat metrics016 and -services as the same. 
We still need to keep metric08 as this is what is in openshift 3.3 and earlier

josejulio:
I'll fix that and post a PR soon.

spec/integration/metric_spec.rb
```diff
@@ -135,6 +135,7 @@
           id1 = SecureRandom.uuid
           id2 = SecureRandom.uuid
           id3 = SecureRandom.uuid
+          id4 = SecureRandom.uuid
 
           ids = [id1, id2, id3]
           bindings = { id1: id1, id2: id2, id3: id3 }
```
josejulio: Could you please add id4 to this hash (`bindings`)?
It will help VCR do its replace-template-magic-thing.

jshaughn:
@josejulio, This dumb PR is finally green :)  Please make it go away unless you see something else that needs to be fixed.  Thanks, Jay

josejulio:
Thank you!

PR #161, opened 2016-10-13
Allow to list tags in hawkular-metrics

rubenvp8510:
Need to add specs.

### DIFF
--- a/lib/hawkular/metrics/metric_api.rb
+++ b/lib/hawkular/metrics/metric_api.rb
@@ -69,6 +69,18 @@ module Hawkular::Metrics
       http_post(path, data)
     end
 
+    # Fetch all tags for metrics definitions
+    # @return [Hash{String=>String}]
+    def tags
+      tags = []
+      http_get('/metrics/').map do |g|
+        g['tags'].map do |k, v|
+          tags << { k => v }
+        end unless g['tags'].nil?
+      end
+      tags.uniq!
+    end
+
     # Base class for accessing metric definition and data of all
     # types (counters, gauges, availabilities).
     class Metrics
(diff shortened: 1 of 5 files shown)


lib/hawkular/metrics/metric_api.rb
```diff
@@ -69,6 +69,18 @@ def query_stats(gauge_ids: [], counter_ids: [], avail_ids: [], rates: false, sta
       http_post(path, data)
     end
 
+    # Fetch all tags for metrics definitions
+    # @return [Hash{String=>String}]
+    def tags
+      tags = []
+      http_get('/metrics/').map do |g|
+        g['tags'].map do |k, v|
```
pilhuhn, line 77: Can you please check with e.g. @burmanm about tags with a list of values ?

lib/hawkular/metrics/metric_api.rb
```diff
@@ -69,6 +69,18 @@ def query_stats(gauge_ids: [], counter_ids: [], avail_ids: [], rates: false, sta
       http_post(path, data)
     end
 
+    # Fetch all tags for metrics definitions
+    # @return [Hash{String=>String}]
+    def tags
```
pilhuhn, line 74: Would it make sense to allow limiting by key - e.g. 'only return tags that start with app:' ?

lib/hawkular/metrics/metric_api.rb
burmanm: We don't currently support single tag with multiple values. It's currently a solution dependant on how to do this, for example in Openshift integration we do with comma separated list. If there's a need for such in this solution, could you create a JIRA ticket about it? We could then look at implementing it in some sane way..

pilhuhn:
On 17 Oct 2016, at 9:05, Michael Burman wrote:

> We don't currently support single tag with multiple values.

This is ok then. I just saw the openshift thing.
Perhaps we should support this here as well (?)

pilhuhn:
Thanks @rubenvp8510

PR #166, opened 2016-10-19
Add list of changes and bump the version to 2.8.0

pilhuhn:
@josejulio can you check?

### DIFF
--- a/CHANGES.rdoc
+++ b/CHANGES.rdoc
@@ -3,6 +3,17 @@
 This document describes the relevant changes between releases of the
 _hawkular-client_ project.
 
+=== V 2.8.0
+
+* Allow to list tags in Hawkular-Metrics
+* Adding support for Hawkular-Admin-Token via admin_token
+* "ackBy no longer exists on alerts" -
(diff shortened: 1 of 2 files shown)


CHANGES.rdoc
```diff
@@ -3,6 +3,17 @@
 This document describes the relevant changes between releases of the
 _hawkular-client_ project.
 
+=== V 2.8.0
+
+* Allow to list tags in Hawkular-Metrics
+* Adding support for Hawkular-Admin-Token via admin_token
+* "ackBy no longer exists on alerts" -
```
josejulio: ackBy does no longer exists on the backend but ackBy still exists on the client 
  https://github.com/jshaughn/hawkular-client-ruby/blob/89674838f086eb7c594a8c877dbd486c82939fc3/lib/hawkular/alerts/alerts_api.rb#L547
Though I don't mind to leave it as is.
Also there is an extra '-' at the end, do you mind removing it?

josejulio:
LGTM

PR #167, opened 2016-10-28
Fixes get_entity call.

josejulio:
`CanonicalPath.to_s` starts with a '/' thus the url has
consecutive slashes and Hawkular fails to parse that
path.


jshaughn:
LGTM

PR #171, opened 2016-11-17
Update inventory client to match REST API: metricDataType

jotak:
Metric type field is now named "metricDataType".
This change has been live for a while, see https://github.com/hawkular/hawkular-inventory/pull/283 (0.18.0.Final). Previous naming still works as deprecated but might be removed soon

### DIFF
--- a/lib/hawkular/inventory/entities.rb
+++ b/lib/hawkular/inventory/entities.rb
@@ -91,7 +91,7 @@ module Hawkular::Inventory
 
     def initialize(type_hash)
       super(type_hash)
-      @type = type_hash['type']
+      @type = type_hash['metricDataType']
       @unit = type_hash['unit']
       @collection_interval = type_hash['collectionInterval']
     end
@@ -107,7 +107,7 @@ module Hawkular::Inventory
 
     def initialize(metric_hash)
       super(metric_hash)
-      @type = metric_hash['type']['type']
+      @type = metric_hash['type']['metricDataType']
       @type_path = metric_hash['type']['path']
       @type_id = metric_hash['type']['id']
       @unit = metric_hash['type']['unit']
--- a/lib/hawkular/inventory/entities.rb
+++ b/lib/hawkular/inventory/entities.rb
@@ -75,7 +75,7 @@ module Hawkular::Inventory
 
   # Fields that are common for MetricType and Metric
   module MetricFields
-    # @return [String] GAUGE, COUNTER, etc.
+    # @return [String] gauge, counter, etc.
     attr_reader :type
     # @return [String] metric unit such as NONE, BYTES, etc.
     attr_reader :unit
--- a/lib/hawkular/inventory/inventory_api.rb
+++ b/lib/hawkular/inventory/inventory_api.rb
@@ -243,14 +243,14 @@ module Hawkular::Inventory
     #   result by a filter to only return a subset. The
     # @param [String] resource_path Canonical path of the resource.
     # @param [Hash{Symbol=>String}] filter for 'type' and 'match'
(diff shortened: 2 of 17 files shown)


jotak:
@metlos can you also have a look? Nothing unexpected?

josejulio:
@jotak I'm looking why the tests are failing. Thought it doesn't seem related to your changes.

josejulio:
@jotak 
Error 1: https://travis-ci.org/hawkular/hawkular-client-ruby/jobs/176631602#L416
and Error 4: https://travis-ci.org/hawkular/hawkular-client-ruby/jobs/176631602#L489
It seems that `metricDataType` are all in lowercase, while `type` are in upper case. If this is the expected scenario, could you please update the hawkular-client-ruby code to use a non-sensitive case comparison? 
https://github.com/hawkular/hawkular-client-ruby/blob/f5dd19947bd05d022ef3f2633f6008e8466eb239/lib/hawkular/inventory/inventory_api.rb#L554

Error 2: https://travis-ci.org/hawkular/hawkular-client-ruby/jobs/176631602#L437
and 3 https://travis-ci.org/hawkular/hawkular-client-ruby/jobs/176631602#L463
Again, we are using uppercase, but it seems that it's expecting lower case (even though the error says that it should be on uppercase)
`Can not construct instance of org.hawkular.inventory.api.model.MetricDataType from String value 'GAUGE': value not one of declared Enum instance names: ****[GAUGE, AVAILABILITY, COUNTER, COUNTER_RATE]****`
Could you check if the lowercase is intended? if so, maybe update the error message. 

Thanks.

jotak:
@josejulio thanks for the analysis ; I'm surprised because the values shouldn't have changed, in inventory the deprecated "type" getter was just redirecting to "metricDataType", so it should have been transparent. I'm having a look at it.

jotak:
@josejulio you were right, the deserialization in inventory follows a different process, I didn't see it. I'll fix it.

jotak:
@josejulio , any objection if I change all the metric types in lower case?

josejulio:
@jotak Not sure if that would affect other clients also. If changing to lower case could you please notify to other clients? 
Also please update the error message, as it seems to asks for upper case types.

 `Can not construct instance of org.hawkular.inventory.api.model.MetricDataType from String value 'GAUGE': value not one of declared Enum instance names: ****[GAUGE, AVAILABILITY, COUNTER, COUNTER_RATE]****`

jotak:
@josejulio so, I confirm that inventory metricDataType is low case and we should keep it that way (it's more consistent with metrics, and there's also other reasons for that...). I've updated this PR is this way.

I'm checking the usage in the other clients.

On inventory side, I have opened a ticket to fix the error message: https://issues.jboss.org/browse/HWKINVENT-208 . It will be fixed when we will finally remove the deprecated field (the pull request is ready, but we will probably wait a little bit more).

josejulio:
@jotak Could you rebase to fix the errors?

jotak:
done

josejulio:
Would this fail for clients that are using things like:
@client.list_metrics_for_resource(wild_fly.path, type: 'GAUGE', match: 'Metrics~Heap') ?

I'm a bit concerned that bugs of this kind might popup on the manageiq integration.
(i.e. currently using uppercase strings)

jotak:
Yes you're right, this is a deprecated usage.
I need to see how I can run the manageiq test suite against my own inventory server...

josejulio:
@jotak Let me know if you need help with this one. I suppose I could run MIQ tests (using this PR) and latest hawkular-services.

jotak:
Hi, there's no need to spend time on this as it's probably going to be obsolete anyway. I keep it opened for now, but it's probably going to be dismissed.

PR #172, opened 2016-11-22
Making sure we compare 2 strings and not symbol with string.

jkremser:
This PR fixes this issue: https://bugzilla.redhat.com/show_bug.cgi?id=1391694#c5

If the operation is called like [this](https://github.com/ManageIQ/manageiq/blob/67f80275ef8224626dfb65f8496c0b62f2bcb3a0/app/models/manageiq/providers/hawkular/middleware_manager.rb#L171) (the operation name is given by a symbol and not as a string), it is performed correctly, but the error call back is called.


josejulio:
@Jiri-Kremser  could you rebase? some errors are because the tests ran before the agent finished.

josejulio:
LGTM

PR #173, opened 2016-11-22
Fixes dependency requeriment for test suite on Ruby 2.3.0

josejulio:
Locked action-pack dependency on latest 4.
When testing on a live server, increases wait time from 20s to 120s to wait for the hawkular wildfly agent.


josejulio:
@rubenvp8510 please review

rubenvp8510:
This looks good to me.

PR #177, opened 2016-11-24
Added interval field to conditions.

### LINKED ISSUE #174: Add 'interval' field of alert conditions
[13:56:08]  <pilhuhn>	lponce for the MISSING condition, the rest api talks about a field 'interval' - is that right? If so would that need a new field in the condition of the ruby client?
[13:57:21]  <lponce>	pilhuhn, yes, let me remember the API
[13:58:00]  <lponce>	pilhuhn, http://www.hawkular.org/docs/rest/rest-alerts.html#MissingCondition yes, an interval field is required

rubenvp8510:
Fixes #174


josejulio:
LGTM

PR #179, opened 2016-11-30
Added an option to query metrics based on tag only (regardless to type)

### LINKED ISSUE #175: Add an option to query metrics based on tag only (regardless to type)
```ruby
Hawkular::Metrics::Client.new(entrypoint, credentials, options).strings.get_data_by_tags(:miq_metric => true)
```

will work, but i need it to work without the strings section.
cc @pilhuhn @zgalor @simon3z.

rubenvp8510:
Fixes #175

Also, added buckets parameter to get_data_by_tags method.

### DIFF
--- a/lib/hawkular/metrics/metric_api.rb
+++ b/lib/hawkular/metrics/metric_api.rb
@@ -12,6 +12,25 @@ module Hawkular::Metrics
       array
     end
 
+    def tags_param(tags)
+      tags.map { |k, v| "#{k}:#{v}" }.join(',')
+    end
+
+    # Retrieve all types of metrics datapoints by tags
+    # @param tags [Hash]
+    # @param buckets [Integer] optional number of buckets
+    # @param bucketDuration [String] optional interval (default no aggregation)
+    # @param starts [Integer] optional timestamp (default now - 8h)
+    # @param ends [Integer] optional timestamp (default now)
+    # @return [Array[Hash]] datapoints
+    def data_by_tags(tags, buckets: nil, bucketDuration:nil, start:nil, ends: nil)
+      data = {
+        tags: tags_param(tags), buckets: buckets, bucketDuration: bucketDuration, start: start, end: ends
+      }
+
+      http_post('metrics/stats/query', data)
+    end
+
     # Return version and status information for the used version of Hawkular-Metrics
     # @return [Hash{String=>String}]
     #         ('Implementation-Version', 'Built-From-Git-SHA1', 'Status')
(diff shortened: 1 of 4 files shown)


lib/hawkular/metrics/metric_api.rb
```diff
@@ -191,19 +203,15 @@ def raw_data(ids, starts: nil, ends: nil, limit: nil, order: nil)
...
+      def get_data_by_tags(tags, starts: nil, ends: nil, bucketDuration: nil, buckets:nil)
+        params = { tags: @client.tags_param(tags), start: starts,
+                   end: ends, bucketDuration: bucketDuration, buckets: buckets }
         path = "/#{@resource}/"
         @legacy_api ? path << 'data/?' : path << 'stats/?'
         resp = @client.http_get(path + encode_params(params))
         resp.is_a?(Array) ? resp : [] # API returns no content (empty Hash) instead of empty array
       end
 
-      def tags_param(tags)
```
josejulio, line 203: What is the reason to move it above ?

rubenvp8510: I need to use it at the level of the Client class. so I moved it to that class in order to reuse the code.

lib/hawkular/metrics/metric_api.rb
```diff
@@ -12,6 +12,18 @@ def default_timestamp(array)
       array
     end
 
+    def tags_param(tags)
+      tags.map { |k, v| "#{k}:#{v}" }.join(',')
+    end
+
+    def get_all_data_by_tags(tags, buckets: nil, bucketDuration:nil, start:nil, ends: nil)
```
josejulio: Could we use data_by_tags instead? It seems to be more ruby-like.

lib/hawkular/metrics/metric_api.rb
```diff
@@ -191,19 +203,15 @@ def raw_data(ids, starts: nil, ends: nil, limit: nil, order: nil)
       # @param bucketDuration [String] optional interval (default no aggregation)
       # @return [Array[Hash]] datapoints
       # @see #push_data #push_data for datapoint detail
-      def get_data_by_tags(tags, starts: nil, ends: nil, bucketDuration: nil)
-        params = { tags: tags_param(tags), start: starts,
-                   end: ends, bucketDuration: bucketDuration }
+      def get_data_by_tags(tags, starts: nil, ends: nil, bucketDuration: nil, buckets:nil)
```
josejulio, line 214: Could you document the buckets param?

lib/hawkular/metrics/metric_api.rb
rubenvp8510: You're right, I'll rename this method.

josejulio: While you are at it, could you also document the method and its params ?

lib/hawkular/metrics/metric_api.rb
josejulio: Also, maybe (in a separate-PR) we should deprecate this `get_data_by_tags` and call it just `data_by_tags`... (using an alias).

josejulio:
LGTM

simon3z:
@yaacov this may be interesting for you as well.

PR #182, opened 2017-01-06
Ensures consistent behavior with returned resource when creating a re…

### LINKED ISSUE #180: @inv_client.create_resource does not return path when resource already exists
This goes 

```
def create_resource_under_resource
...
      rescue HawkularException => error
        # 409 We already exist -> that is ok
        raise unless error.status_code == 409
      end
      Resource.new(res)
```

In case of the 409, we still return a `Resource`, but constructed from the input, which does not contain the path.

The path needs to be obtained by clients like this:

```
resource = @inv_client.create_resource  type_path, data['r'], data['r']
  if resource.path.nil?
    resources = @inv_client.list_resources_for_type type_path, true
    resource = resources.first
  end
```

We should perhaps internalise getting the path into `create_resource` to have a consistent behaviour in both cases.

josejulio:
…source that already exists.

Fixes #180

### DIFF
--- a/lib/hawkular/inventory/entities.rb
+++ b/lib/hawkular/inventory/entities.rb
@@ -193,6 +193,15 @@ module Hawkular::Inventory
       CanonicalPath.new(hash)
     end
 
+    # Adds a resource to the path.
+    # @return CanonicalPath referring to resource_id using current as its ancestor.
+    def to_resource(resource_id)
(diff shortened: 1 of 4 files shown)


josejulio:
@rubenvp8510 could you please review?

rubenvp8510:
LGTM

PR #183, opened 2017-02-01
update tests to be compatible with latest h-alerts (1.5.1) in h-services

jshaughn:
- new field ActionDefinition.global
- the alerts email plugin is no longer explicitly deployed, it is now
  embedded in hawkular-metrics.ear.

@josejulio Please review.  Note, I haven't really tested these changes (I'm not sure how).  The aren't expected to pass unless run against the Services PR with the updates.  Let me know if there is something I can do to validate.

### DIFF
--- a/spec/integration/alerts_spec.rb
+++ b/spec/integration/alerts_spec.rb
@@ -192,7 +192,7 @@ module Hawkular::Alerts::RSpec
 
       ret = @client.get_action_definition 'email'
       expect(ret.size).to be(1)
-      expect(ret['email'].size).to be(7)
+      expect(ret['email'].size).to be(8)
 
       expect { @client.get_action_definition '-does-not-exist-' }
         .to raise_error(Hawkular::BaseClient::HawkularException)
--- a/spec/integration/operations_spec.rb
+++ b/spec/integration/operations_spec.rb
@@ -190,14 +190,14 @@ module Hawkular::Operations::RSpec
 
         it 'Restart should be performed and eventually respond with success' do
           wf_server_resource_id = 'Local~~'
-          alerts_war_resource_id = 'Local~%2Fdeployment%3Dhawkular-alerts-actions-email.war'
+          status_war_resource_id = 'Local~%2Fdeployment%3Dhawkular-status.war'
           path = CanonicalPath.new(tenant_id: @tenant_id,
                                    feed_id: @feed_id,
-                                   resource_ids: [wf_server_resource_id, alerts_war_resource_id])
+                                   resource_ids: [wf_server_resource_id, status_war_resource_id])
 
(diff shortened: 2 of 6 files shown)


josejulio:
The changes look good. 

I guess that you already tested them because your cassettes are updated (unless you manually edited).

Anyway we should merge this one once Hawkular-services related PR is already merged.

In case you haven't been able to test on the services-PR
You can execute:
`VCR_OFF=1 bundle exec rspec spec/integration/alerts_spec.rb:188`
and
`VCR_OFF=1 bundle exec rspec spec/integration/operations_spec.rb:191`

while having Hawkular services running on `localhost:8080`

jshaughn:
@josejulio Yes, I was able to test against my hServices branch locally and update the cassettes via VCR_UPDATE=1.  Things are passing, so, as you say, I think we should merge the services PR and then this one.

jshaughn:
@josejulio or @pilhuhn, The hServcices PR was merged so please merge this as well. Thanks!

pilhuhn:
Shouldn't this then succeed against a live server? And also without server?

jshaughn:
Yes, it should succeed without a live server and should now succeed against hServices master.  I don't think it has been run against the latest live server, let me restart the test then...

jshaughn:
@pilhuhn @josejulio I'm not exactly sure what hServices this runs against, the failure seems to me that it still runs against an older hServices image (pilhun:latest?).  I don't know, I'm fairly sure the specs are now compatible with what is in master.  Perhaps you guys can figure out what's wrong and whether this should be merged.  I'm happy to help, just not sure what to do.

josejulio:
@jshaughn is indeed trying to use pilhun:latest, see https://github.com/hawkular/hawkular-client-ruby/blob/master/docker-compose.yml#L2
Does H-services publish one too?

pilhuhn:
Well, pilhuhn:latest is the latest release. Even if there would be an 
official one, we should still test the ruby client against a release 
version and not arbitrary master, as otherwise we would have too many 
moving parts


Am 2. Februar 2017 5:44:50 nachm. schrieb Josejulio Martínez 
<notifications@github.com>:

jshaughn:
@pilhuhn @josejulio OK, then if I understand this correctly, this PR needs to wait until the next hServices release goes out an pilhuhn:latest is updated.  At that time the PR should pass all tests.  For now it will only pass the cassette test using the cassettes supplied in the PR.  I will leave the DNM label in place until the next release goes out.  Thanks, let me know if this is incorrect.

josejulio:
That seems to be correct, Thanks!

pilhuhn:
This now works with hawkular-services 0.31

PR #184, opened 2017-02-07
Some Code Refactors

cfcosta:
This PR is done and can be merged :D

Those are the changes:

* Refactored HawkularUtilsMixin
  * Changes the array of gsubs to not modify the object in-place but generate a new one, since that's already the behaviour Ruby does by default (changing a duplicate and `String.replace`-ing it).
  * Renames HawkularUtilsMixin to avoid both polluting the main scope and having a "type" in the name.
* Extract ENV configuration and Loggers into their own files
  * The idea behind this change is to avoid putting fetching data from the environment where it doesn't belong (which makes change ENV keys a shotgun surgery), and also removing `puts` statements from the classes, making it easier to log into a file if necessary.
* Removing the file expansion for tests on requires
  * `require_relative` exists since Ruby 1.9.2, and makes the code cleaner.

If there are any other pain points on the codebase that make you feel sad everytime that you need to change them, can you also pinpoint those? I might be able to fix them or at least make it better.


josejulio:
Are you going to put every used ENV on `lib/hawkular/env_config.rb` ?

cfcosta:
@josejulio only the ones that are actually used in development/production, I do not plan to add test-only envs there.

josejulio:
> This is not done yet, because for some reason changing OperationsClient.log_message in any way (both removing the puts and the if to use the logger and just using EnvConfig instead of ENV cause many tests in the build to fail), and I need some ideas on how to do it. Any feedback on why this might be happening?

What tests are failing? Travis seems to be happy on both live server and cassettes

cfcosta:
Fixed the problem I reported, turns out that the context changes on the ws block, so you can not access instance variables or methods, but the errors are swallowed on the tests, returning nil when it should raise an error :\

@josejulio can you do a review please?

lib/hawkular/logger.rb
```diff
@@ -0,0 +1,16 @@
...
+module Hawkular
+  class Logger
+    def initialize(file = STDOUT)
+      @logger = ::Logger.new(file)
+    end
+
+    def log(message, priority = :info)
+      return unless EnvConfig.log_response?
+
+      @logger.send(priority, message)
```
josejulio: Not really sure if it would be more readable with `@logger.send(priority, message) if EnvConfig.log_response?`. Less code surely, but more readable? WDYT?

lib/hawkular/operations/operations_api.rb
```diff
@@ -1,6 +1,7 @@
 require 'hawkular/base_client'
 require 'websocket-client-simple'
 require 'json'
+require 'pry'
```
josejulio: I suppose this is used for debug? If so please remove it

cfcosta:
@josejulio fixed both :D

lib/hawkular/operations/operations_api.rb
```diff
@@ -30,16 +32,14 @@ module Hawkular::Operations
...
     # helper for parsing the "OperationName=json_payload" messages
     class WebSocket::Frame::Data
       def to_msg_hash
         chunks = split('=', 2)
-        {
-          operationName: chunks[0],
-          data: JSON.parse(chunks[1])
-        }
+
+        { :operationName => chunks[0], :data => JSON.parse(chunks[1]) }
```
josejulio: Could you please tell why changing to this syntax? Also our rubocop is configured to use the Ruby: 1.9 syntax over hash rocket.

lib/hawkular/operations/operations_api.rb
```diff
@@ -62,39 +62,53 @@ def to_msg_hash
...
-
-      unless ep.nil?
-        uri = URI.parse ep.to_s
+      args = {
+        :credentials => {},
+        :options => {},
+        :wait_time => 0.5,
+        :use_secure_connection => false,
+        :entrypoint => nil
+      }.merge(args)
```
josejulio: Please update to use the Ruby 1.9 syntax

cfcosta: @josejulio oh, ok, for some reason I thought this project was using the same ruleset as ManageIQ.

josejulio: I though the same when starting :smile:

josejulio:
> @josejulio fixed both :D

Thank you. Rubocop is complaining about this:
```
lib/hawkular/operations/operations_api.rb:83:7: W: Useless assignment to variable - ws_options.
      ws_options = {}
      ^^^^^^^^^^
```

https://travis-ci.org/hawkular/hawkular-client-ruby/jobs/205872877#L366

josejulio:
Thanks @cfcosta !

abonas:
everyone, in the future, please:
1. do not merge commits that start with WIP
2. keep in mind that the description of this PR with several commits is not reflected in git log, hence some explanations are "lost". this is a very big PR with unrelated changes. breaking it to smaller separate PRs with a clear description would be much better.


@Jiri-Kremser please elaborate what do you mean [here](https://github.com/hawkular/hawkular-client-ruby/pull/184#commitcomment-21291272)

josejulio:
I suppose @Jiri-Kremser means the refactor of 'HawkularUtilsMixin'. 
It was reintroduced and marked as deprecated a follow up here https://github.com/hawkular/hawkular-client-ruby/pull/194

jkremser:
Yes as JoseJulio said. It was comment on the commit. I didn't know that GitHub is smart enough to show this comment also in the PR that contains that commit.

PR #186, opened 2017-02-10
Operation add datasource and remove datasource fails if running on th…

josejulio:
…e newest hawkular-services container

We now detect (currently via env variable) if we are running in one and expects those tests to fail.


josejulio:
I'll try to get the server [immutable flag](https://github.com/hawkular/hawkular-agent/pull/286) instead of setting a value.

josejulio:
@pilhuhn This tries to fetch the "In Container" property from the agent to determine how to behave on a couple of operation specs. If said property isn't found, we default to "In Container" = false.

pilhuhn:
I'll merge that now. In-Container flag will be provided for container builds once https://github.com/hawkular/hawkular-services/pull/122 is merged.

josejulio:
There seems to be new errors there, I'll check it and follow this up.

PR #187, opened 2017-02-20
HAWKULAR-1154 - Supports [Update Collection Intervals] command.

josejulio:


josejulio:
We need to wait for a hawkular-services release with `Update Collection Intervals`  command included.

jshaughn:
I believe 0.32 services has the request/response already, but your right, we need an updated agent for end-to-end.

josejulio:
@rubenvp8510 please review

PR #188, opened 2017-02-23
Lazy load endpoints.

josejulio:
Lazy loading the endpoints. This will keep our users for having to do stuff like https://github.com/ManageIQ/manageiq/pull/13814

@cfcosta could you please review ?

// cc @pilhuhn

### DIFF
--- a/lib/hawkular/hawkular_client.rb
+++ b/lib/hawkular/hawkular_client.rb
@@ -14,31 +14,15 @@ module Hawkular
       hash[:options] ||= {}
       fail 'no parameter ":entrypoint" given' if hash[:entrypoint].nil?
       @state = hash
-
-      @inventory = Inventory::InventoryClient.create(entrypoint: "#{hash[:entrypoint]}/hawkular/inventory",
-                                                     credentials: hash[:credentials],
-                                                     options: hash[:options])
-
-      @metrics = Metrics::Client.new("#{hash[:entrypoint]}/hawkular/metrics",
-                                     hash[:credentials],
-                                     hash[:options])
-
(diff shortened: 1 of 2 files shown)


lib/hawkular/hawkular_client.rb
```diff
@@ -47,13 +31,37 @@ def method_missing(name, *args, &block)
       delegate_client.__send__(method, *args, &block)
     end
 
+    def inventory
+      @inventory ||= Inventory::InventoryClient.create(entrypoint: "#{@state[:entrypoint]}/hawkular/inventory",
```
cfcosta, line 35: Why does `Metrics` has a different class name? Can we change the others to be like it? I like it better than to have `Inventory::Inventory` and `Alerts::Alerts`.

cfcosta: Also the initialization is different on `Inventory` but not on the others.

cfcosta (approved):
LGTM, there are some improvements that can be done but they might be out of scope of this PR.

lib/hawkular/hawkular_client.rb
josejulio: I'm not sure, but at this point i don't want to break stuff like this: https://github.com/ManageIQ/manageiq/pull/13814/files#diff-7873cbd67d037fe54151ac4d277ccdd4R44

cfcosta: You can have this at the end of the file and keep it compatible:

```ruby
Inventory::InventoryClient = Inventory::Client
```

josejulio: Will open a separate PR to uniform it.

josejulio:
@rubenvp8510 could you please merge?

PR #189, opened 2017-02-24
Uniform Client names.

josejulio:
Added backwards compatibility for old names, but marked them as
deprecated.

### DIFF
--- a/lib/hawkular/alerts/alerts_api.rb
+++ b/lib/hawkular/alerts/alerts_api.rb
@@ -13,7 +13,7 @@ module Hawkular::Alerts
   #   http://localhost:8080/hawkular/alerts
   # @param credentials [Hash{String=>String}] Hash of username, password, token(optional)
   # @param options [Hash{String=>String}] Additional rest client options
-  class AlertsClient < Hawkular::BaseClient
+  class Client < Hawkular::BaseClient
     def initialize(entrypoint, credentials = {}, options = {})
       entrypoint = normalize_entrypoint_url entrypoint, 'hawkular/alerts'
       @entrypoint = entrypoint
@@ -559,4 +559,7 @@ module Hawkular::Alerts
       super(event_hash)
     end
   end
+
+  AlertsClient = Client
+  deprecate_constant :AlertsClient if self.respond_to? :deprecate_constant
 end
--- a/lib/hawkular/hawkular_client.rb
+++ b/lib/hawkular/hawkular_client.rb
@@ -32,9 +32,9 @@ module Hawkular
     end
 
     def inventory
-      @inventory ||= Inventory::InventoryClient.create(entrypoint: "#{@state[:entrypoint]}/hawkular/inventory",
-                                                       credentials: @state[:credentials],
-                                                       options: @state[:options])
+      @inventory ||= Inventory::Client.new("#{@state[:entrypoint]}/hawkular/inventory",
+                                           @state[:credentials],
+                                           @state[:options])
     end
 
     def metrics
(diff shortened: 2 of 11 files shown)


josejulio:
@rubenvp8510 @cfcosta  please review when you have some time.

lib/hawkular/alerts/alerts_api.rb
```diff
@@ -559,4 +559,7 @@ def initialize(event_hash)
       super(event_hash)
     end
   end
+
+  AlertsClient = Client
+  deprecate_constant :AlertsClient
```
cfcosta: Didn't know about this, nice!

lib/hawkular/hawkular_client.rb
```diff
@@ -66,7 +66,7 @@ def tokens
 
     # this is in a dedicated method, because constructor opens the websocket connection to make the handshake
     def init_operations_client
-      Operations::OperationsClient.new(entrypoint: @state[:entrypoint],
+      Operations::Client.new(entrypoint: @state[:entrypoint],
```
cfcosta: What do you think about also make the arguments uniform? We could have a new method like `#create` so we not break compatibility.

cfcosta (approved):
LGTM!

lib/hawkular/hawkular_client.rb
josejulio: It just worries me that Inventory has a `#create` method but is being used like Operation's constructor. 
Do you think it would confuse users? Maybe not.. but though on asking.

lib/hawkular/alerts/alerts_api.rb
josejulio: `deprecate_constant` doesn't seem to exist on our base testing version. Will fix it later.

lib/hawkular/hawkular_client.rb
cfcosta: I don't think so as long as we're using it only internally (and we can deprecate using `#new` directly), the user is not expected to be running `Operations::Client` directly, is it?

josejulio: It *could* but not sure if anyone is doing it. 
Anyway i'll use a #create method.

pilhuhn:
When you make changes - especially to the metrics sub-client, but not only. 
Make sure to sync with @simon3z and his team, as the ruby-client is not only used by the Middleware provider.

simon3z:
> Make sure to sync with @simon3z and his team, as the ruby-client is not only used by the Middleware provider.

@yaacov @moolitayer do we have spec tests that will fail in ManageIQ when this gem will be updated?
Will those failures give us (and @pilhuhn and @josejulio) enough feedback on what should be changed in order not to break hawkular metrics collection for OpenShift?

yaacov:
@pilhuhn thanks, @simon3z in _metrics_ we already use `Client`
https://github.com/hawkular/hawkular-client-ruby/blob/master/lib/hawkular/metrics/metrics_client.rb#L21
so this PR will not change the _metrics_ code.

We have VCR specs for `inventory` and `metrics` that will fail if the ruby client change it's API interface.

moolitayer:
@simon3z The datawarehouse manager also has specs in place that should fail in event collection.
(Also VCR)

We use AlertsClient [now](https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/hawkular/datawarehouse_manager.rb#L46)

I will send an update, this change seems backward compatible in the mean time.

Anyway clients are much more consistent now.

josejulio:
I tried to keep a backwards compatible API, so at most you would get some deprecation warnings.

rubenvp8510:
This PR passed all Travis test, I think it is ready for be merged.

PR #190, opened 2017-02-27
HAWKULAR-1205 - Adds the ability to assign tags to existing alerts.

josejulio:


josejulio:
@rubenvp8510 could you please review?

lucasponce:
@josejulio could you add the counterpart DELETE tag endpoint on this PR too ?
http://www.hawkular.org/docs/rest/rest-alerts.html#DELETE__tags

Then we will cover the whole use case which is basically to add/remove custom tags from ManageIQ.

Thanks !

josejulio:
> @josejulio could you add the counterpart DELETE tag endpoint on this PR too ?
> http://www.hawkular.org/docs/rest/rest-alerts.html#DELETE__tags
> 
> Then we will cover the whole use case which is basically to add/remove custom tags from ManageIQ.
> 
> Thanks !

Sure

josejulio:
@moolitayer Is this what you need?

moolitayer:
Looks exactly like what I need. Thanks @josejulio!

ps @josejulio @lucasponce we would need a new version of the client and an update in Miq,
I can send the later just let me know
Thanks again

PR #191, opened 2017-02-27
Use the official docker repository from hawkular.

josejulio:


josejulio:
Needs to mark some other methods as 'when running in container expect it to fail'

PR #193, opened 2017-03-02
Add an entrypoint for requires of the gem

cfcosta:
Currently you have to add a `require: ` flag when you add the gem on the `Gemfile`. This PR fixes the problem.


josejulio:
LGTM

PR #194, opened 2017-03-13
Fix old usages of internal constants

cfcosta:
This PR re-adds the `HawkularUtilsMixin` constant (I wasn't aware it was being used directly on ManageIQ), and also adds tests for documenting planned deprecations.


jkremser:
lgtm

PR #196, opened 2017-03-17
Inventory on metrics

jotak:
Read inventory from String metrics instead of hawkular-inventory
Note, many functions have been removed for the time being, as not used in MIQ. We may add them back later.

WIP: tests not refactored yet, thus failing

This PR takes place in the context of replacing hawkular-inventory with self-managed inventory by hawkular-agent and this ruby client, using hawkular-metrics/C* as backend. It's related to this agent-side PR: https://github.com/hawkular/hawkular-agent/pull/303

In the new architecture, the inventory structure is stored as compressed JSON blobs, eventually chunked, in hawkular-metrics as string-typed metrics. The temporal value of the metric is not relevant: only the last datapoint would be considered (but this is not true in the implementation details, because the blobs are split in several chunks, which are stored in several datapoints - which means that to get the current state of an inventory, we fetch the entire relevant metric datapoints)

There is one metric per resource type, metric type and root resource (and partitioned by feed id). All child resources are stored within their parent root structure.

Metrics are named with the following pattern:
`inventory.#{feedid}.#{type}.#{id}`

`#{type}` being one of `rt` (resource type), `mt` (metric type), `r` (root resource), thus keeping the same terminology as before with inventory.

In addition, each metric is tagged with the following:
`{module: 'inventory', feed: #{feedid}, type: #{type}, id: #{id}}`

Metrics for root resources also provide the `restypes` tag, that contains their resource types and those of all their children, separated with '|' ; for instance: `restypes:|Wildfly Server|Datasource|etc.|`. That allows to keep providing fast access to common queries such as "search all resources of a given type".


jotak:
DO NOT MERGE YET

jotak:
@Jiri-Kremser here's a first commit. I need to update all tests now

As discussed on the ml, I've removed many functions as they don't seem to be used in MIQ. That makes the client very tight to MIQ now, but we may add them back later

jotak:
Also, I'm a total noob in ruby so don't hesitate to point any code style issue (even if Rubocop helped me a bit here)

lib/hawkular/inventory/entities.rb
```diff
@@ -229,6 +231,18 @@ def to_s
       ret
     end
 
+    def to_metric_name
+      fail 'Missing feed_id' if @feed_id.nil?
```
cfcosta: `fail 'Missing feed_id' unless @feed_id`

lib/hawkular/inventory/entities.rb
```diff
@@ -229,6 +231,18 @@ def to_s
...
+      if @resource_type_id.nil? && @metric_type_id.nil? && (@resource_ids.nil? || @resource_ids.empty?)
+        fail 'Expecting either resource type, metric type or resource'
+      end
+      ret = "inventory.#{@feed_id}"
+      ret += ".rt.#{@resource_type_id}" unless @resource_type_id.nil?
```
cfcosta: `if @resource_type_id`

lib/hawkular/inventory/entities.rb
```diff
@@ -229,6 +231,18 @@ def to_s
...
+      ret += ".mt.#{@metric_type_id}" unless @metric_type_id.nil?
```
cfcosta: `if @metric_type_id`

lib/hawkular/inventory/entities.rb
```diff
@@ -229,6 +231,18 @@ def to_s
...
+      ret += ".r.#{@resource_ids[0]}" unless @resource_ids.nil?
```
cfcosta: `if @resource_ids`

lib/hawkular/inventory/inventory_api.rb
```diff
@@ -80,13 +73,19 @@ def list_resource_types(feed_id = nil)
...
+        limit: 1,
+        order: 'DESC',
+        tags: "module:inventory,type:r,feed:#{the_feed}"
+      )
       to_filter = ret.map do |r|
-        if fetch_properties
-          p = get_config_data_for_resource(r['path'])
-          r['properties'] = p['value']
+        json = extract_metric_json(r)
+        unless json.nil?
```
cfcosta: `if json`

lib/hawkular/inventory/inventory_api.rb
```diff
@@ -136,433 +131,122 @@ def get_config_data_for_resource(resource_path)
...
-      end
-    end
-
-    # Delete the feed with the passed feed id.
-    # @param feed_id Id of the feed to be deleted.
-    def delete_feed(feed_id)
-      the_feed = hawk_escape_id feed_id
-      http_delete("/entity/f;#{the_feed}")
+      raw_hash = get_raw_entity_hash(path)
+      return if raw_hash.nil?
```
cfcosta: `unless raw_hash`

lib/hawkular/inventory/inventory_api.rb
```diff
@@ -136,433 +131,122 @@ def get_config_data_for_resource(resource_path)
...
-        rl = {}
-        rl['otherEnd'] = the_metric.path.to_s
-        rl['name'] = 'incorporates'
-        http_post("/entity/f;#{feed_id}/#{res_path_str}/relationship", [rl])
-      rescue HawkularException => error
-        # 409 We already exist -> that is ok
-        raise unless error.status_code == 409
+    def shorten_resource_type(resource_type)
+      # cf InventoryStructure.class$EntityType
+      case resource_type
```
cfcosta: You can have that as a hash constant and just access it.

lib/hawkular/inventory/inventory_api.rb
```diff
@@ -136,433 +131,122 @@ def get_config_data_for_resource(resource_path)
...
-    #   it will not be used for filtering
-    # @return [Array<Metric>] List of metrics that can be empty.
-    # @example
-    #    # Filter by type and match on metrics id
-    #    client.list_metrics_for_resource(wild_fly, type: 'GAUGE', match: 'Metrics~Heap')
-    #    # Filter by type only
-    #    client.list_metrics_for_resource(wild_fly, type: 'COUNTER')
-    #    # Don't filter, return all metric definitions
-    #    client.list_metrics_for_resource(wild_fly)
-    def list_metrics_for_resource(resource_path, filter = {})
```
jotak: Looks like I've been too brutal in removals. This one must be taken back.

lib/hawkular/inventory/inventory_api.rb
```diff
@@ -102,32 +101,28 @@ def list_resources_for_type(resource_type_path, fetch_properties = false)
...
+          Resource.new(root_hash)
         end
-        Resource.new(r)
       end
     end
 
-    # Retrieve runtime properties for the passed resource
-    # @param [String] resource_path Canonical path of the resource to read properties from.
-    # @return [Hash<String,Object] Hash with additional data
-    def get_config_data_for_resource(resource_path)
```
jotak: rollback this one, used in MIQ

abonas:
> DO NOT MERGE YET

please mark the PR as WIP to indicate it's unmergeable.

lib/hawkular/inventory/inventory_api.rb
```diff
@@ -38,39 +38,32 @@ def self.create(hash)
...
     # List resource types. If no feed_id is given all types are listed
     # @param [String] feed_id The id of the feed the type lives under. Can be nil for feedless types
     # @return [Array<ResourceType>] List of types, that can be empty
-    def list_resource_types(feed_id = nil)
-      if feed_id.nil?
-        ret = http_get('/traversal/type=rt')
-      else
-        the_feed = hawk_escape_id feed_id
-        ret = http_get("/traversal/f;#{the_feed}/type=rt")
+    def list_resource_types(feed_id)
```
cfcosta, line 54: I think this should keep the default value for feed_id.

lib/hawkular/inventory/inventory_api.rb
jotak: yes, that would be pretty much the same... hash versus function, not sure what is the best

lib/hawkular/inventory/inventory_api.rb
jotak: The feed_id was not mandatory in the previous API, but now it is. I checked in MIQ, there should always be a feed id.

lib/hawkular/inventory/inventory_api.rb
cfcosta: Hash, direct access is faster than vtable lookups.

abonas:
Please add a bit more in the PR description about the changes of how the inventory is now read from metrics.

jotak:
@abonas yes, you're right... there's been some discussions about it on the mailing lists, but I'll put more info here.

jotak:
(description updated)

jkremser:
1 big question is in the air (we have already discussed with joel). Can we afford losing the tests+VCRs for the old api? We can create the similar thing we have for metrics, i.e. contexts and put the metadata tags on particular tests if the test make should be run for the context. But it might be an overkill.

lib/hawkular/inventory/entities.rb
```diff
@@ -185,14 +189,38 @@ def self.parse(path)
...
+    # @return a new CanonicalPath based on the current one
+    def rt(resource_type)
+      hash = to_h
+      hash[:resource_type_id] = hawk_escape_id(resource_type)
+      CanonicalPath.new(hash)
+    end
+
+    # Set metric type to the current path
+    # @return a new CanonicalPath based on the current one
+    def mt(metric_type)
```
cfcosta: I don't like these method names, `#rt` and `#mt`. Using `resource_type` and `metric_type` and then aliasing to that name after the definition would be fine though.

lib/hawkular/inventory/inventory_api.rb
```diff
@@ -255,178 +171,60 @@ def list_metrics_for_resource_type(resource_type_path)
...
 
     # List operation definitions (types) for a given resource type
     # @param [String] resource_type_path canonical path of the resource type entity
     # @return [Array<String>] List of operation type ids
     def list_operation_definitions(resource_type_path)
-      parsed_path = CanonicalPath.parse(resource_type_path.to_s)
-      feed_id = parsed_path.feed_id
-      resource_type_id = parsed_path.resource_type_id
-      ots = http_get("/traversal/f;#{feed_id}/rt;#{resource_type_id}/type=ot")
+      path = resource_type_path.is_a?(CanonicalPath) ? resource_type_path : CanonicalPath.parse(resource_type_path)
```
cfcosta: I don't like this. If necessary, `CanonicalPath.parse` should be able to return the caller when it is already a `CanonicalPath`. That would reduce conditionals.

lib/hawkular/inventory/inventory_api.rb
```diff
@@ -255,178 +171,60 @@ def list_metrics_for_resource_type(resource_type_path)
...
+      structures = extract_structures_from_body(response)
       res = {}
-      ots.each do |ot|
-        ot_name = ERB::Util.url_encode ot['name']
-        pts = http_get("/traversal/f;#{feed_id}/rt;#{resource_type_id}/ot;#{ot_name}/d;parameterTypes")
-        ot['parameters'] = pts[0]['value'] unless pts.empty?
-        od = OperationDefinition.new ot
-        res.store od.name, od
+      structures.map { |rt| rt['inventoryStructure'] }
+        .find_all { |rt| (rt.key? 'children') && (rt['children'].key? 'operationType') }
```
cfcosta:
```ruby
.select { |rt| rt['children'] && rt['children']['operationType'] }
```

Would be better, I think.

lib/hawkular/inventory/inventory_api.rb
```diff
@@ -255,178 +171,60 @@ def list_metrics_for_resource_type(resource_type_path)
...
+        .flat_map { |rt| rt['children']['operationType'] }
+        .each do |ot|
+          hash = optype_json_to_hash(ot)
+          od = OperationDefinition.new hash
+          res.store od.name, od
```
cfcosta: `res[od.name] = od`

spec/integration/hawkular_client_spec.rb
```diff
@@ -131,23 +125,24 @@ module Hawkular::Client::RSpec
...
-        expect(feed_list).to include(feed_id2)
-
-        @client.delete_feed feed_id2
-        @hawkular_client.inventory.delete_feed feed_id1
-
-        feed_list = @hawkular_client.inventory_list_feeds
-        expect(feed_list).not_to include(feed_id1)
-        expect(feed_list).not_to include(feed_id2)
-      end
+      # FEED CREATION / DELETION CURRENTLY DISABLED
```
cfcosta: No commented code please, if this is supposed to be enabled later, you can create a commit to remove it and then revert it if necessary.

spec/integration/inventory_spec.rb
```diff
@@ -406,259 +347,19 @@ module Hawkular::Inventory::RSpec
...
-
-        it 'Should reject unknown metric type' do
-          new_feed_id = 'feed_may_exist'
-
-          expect { @client.create_metric_type new_feed_id, 'abc', 'FOOBaR' }.to raise_error(RuntimeError,
-                                                                                            /Unknown type FOOBAR/)
+            .to raise_error(Hawkular::BaseClient::HawkularException, /Resource not found/)
         end
 
         let(:example) do |e|
```
cfcosta, line 357: whaaaat

lib/hawkular/inventory/inventory_api.rb
jotak: True that it would avoid to repeat that small pattern everywhere... however, a "parse" method that could take an already parsed object rather than a string, would be kind of weird. I'd rather add a new private method in the client

lib/hawkular/inventory/entities.rb
jotak: Yep, I'll rename these

lib/hawkular/inventory/inventory_api.rb
jotak: is "select" preferable than "find_all"? It seems to be exactly the same in this case (Array)

lib/hawkular/inventory/inventory_api.rb
jotak: I restored that line from the code before this PR. Is it really equivalent? There might be a reason for having such a writing before

spec/integration/hawkular_client_spec.rb
jotak: Yep, this is related to @Jiri-Kremser comment above: what do we do for obsolete tests? If everybody here is fine about deleting them, I'll do it.

spec/integration/inventory_spec.rb
jotak: ?

lib/hawkular/inventory/inventory_api.rb
cfcosta: There's no difference, `#store` is just an alias to `#[]`: https://docs.ruby-lang.org/en/2.0.0/Hash.html#method-i-store

lib/hawkular/inventory/inventory_api.rb
cfcosta: Not exactly, `CanonicalPath.parse` receives something and returns a CanonicalPath, so if it receives one that's already parsed it just returns it back. That way we don't need to have this pattern everywhere, and only one test is necessary to cover this case. Think String having a `to_s` method.

lib/hawkular/inventory/inventory_api.rb
cfcosta: Well, it's used more, but there's no difference. I'd change just the block, if you prefer `find_all`.

lib/hawkular/inventory/inventory_api.rb
jotak: I'm really not convinced. Why would you want to parse something else than a string (or let's say more generally a character stream)? To me it's a non-sense. String's to_s is also a non-sense, it's just there as a side-effect of most language inheritance design.

I don't say that because I don't want to change my code. I just don't want to flaw the API design for some use case convenience, a function must do just what it's supposed to do, no more, no surprise. Moreover, sometimes the caller knows it has a String, so why performing type checking in that case?

lib/hawkular/inventory/inventory_api.rb
jotak: Is ".key" deprecated or what? Another best practice about not using features? I think Ruby definitely isn't for me :)

lib/hawkular/inventory/inventory_api.rb
jotak: done

lib/hawkular/inventory/inventory_api.rb
cfcosta: Not exactly deprecated, its that having nil as the result and not having the key are equivalents on this case, so checking the key does not cover it.

spec/integration/hawkular_client_spec.rb
josejulio: +1 for having a separate commit to delete them. If we need them we might check that commit (later) and see what we can use.
It will be hard to run these tests if we are calling different endpoints anyway, so i don't see the point.

jotak:
I pushed the updates on tests, with 1 commit for the cassettes update and 1 commit for deletion of obsolete tests.

The current status on tests is:
- I run the tests with the flags `RUN_ON_LIVE_SERVER=1` , `SKIP_SSL_WITHOUT_CERTIFICATE_TEST=1`, `SKIP_SECURE_CONTEXT=1` and `SKIP_V8_METRICS=1`.

- In inventory_spec, they're all green

- Globally, two of them are failing, in operation_spec. "Operation Add JDBC driver should add the driver" fails because my server can't create a file locally to `/opt/jboss/wildfly/modules/foo/barnot_so_random_uuid/main`. I don't know why that, I'm running it in a docker container with volume. The other failing test is a consequence to the first failure.

All other tests are ok.

pilhuhn:
@jotak Failing to add the JDBC driver when the server is in a container is correct. @josejulio did some work to detect that. Does the server in container have the immutable flag (`-Dhawkular.agent.immutable` iirc) set?

jotak:
No, I haven't set that flag. I'll check it.

jkremser:
The hawkfly has the immutable flag set to true [here](https://github.com/hawkular/hawkular-agent/blob/master/docker-dist/Dockerfile#L43). But there is no such a flag for the h-services container. But it's also not possible to add the JDBC datasource on h-services server. iirc we ask if the product = 'Hawkular' in the server properties in MiQ so there must be some similar check in the agent.

@jotak btw. could you pls, do something like:

```diff
diff --git a/docker-compose.yml b/docker-compose.yml
index 14870b2..ba12f8f 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,6 @@
 hawkular:
-  image: "hawkular/hawkular-services:latest"
+  # TODO: put this back to hawkular/hawkular-services:latest once everything is merged in the h-services and the docker image is released
+  image: "jotak/hawkular-services:latest"
   ports:
     - "8080:8080"
     - "8443:8443"
```

So that travis runs the tests against your altered docker image?

docker-compose.yml
```diff
@@ -1,5 +1,6 @@
 hawkular:
-  image: "hawkular/hawkular-services:latest"
+  # TODO: put this back to hawkular/hawkular-services:latest once everything is merged in the h-services and the docker image is released
```
pilhuhn: We must not forget this

simon3z:
@moolitayer @yaacov can you review? Thanks.

yaacov:
> can you review? Thanks.

Thanks, the metrics collection doesn't use inventory , it's ok by me 👍

moolitayer:
> Thanks, the metrics collection doesn't use inventory , it's ok by me 

Agreed, this should not have an impact on containers according to my understanding

lib/hawkular/inventory/inventory_api.rb
```diff
@@ -99,33 +112,34 @@ def list_resources_for_feed(feed_id, fetch_properties = false, filter = {})
...
-          r['properties'] = p['value']
-        end
-        Resource.new(r)
-      end
+      path = CanonicalPath.parse_if_string(resource_type_path)
+      fail 'Feed id must be given' unless path.feed_id
+      fail 'Resource type must be given' unless path.resource_type_id
+
+      # Fetch metrics by tag
+      feed_path = feed_cp(path.feed_id)
```
israel-hdez: I think you need to unescape the `feed_id` or make `feed_cp` use it as-it. It should be already escaped since it comes from a CP. This is making MiQ to fail when WF servers are in domain mode.

jotak:
@israel-hdez my last commit should fix the issue you mentioned by email on weird canonical paths you get with `list_child_resources`. I also rebased from upstream.

lib/hawkular/inventory/inventory_api.rb
jotak: It makes sense to unescape. Done.

PR #197, opened 2017-03-17
[HAWKULAR-1206] add timestamps parameter due to changes in HWKMETRICS…

jsanda:
…-625

### DIFF
--- a/lib/hawkular/metrics/metric_api.rb
+++ b/lib/hawkular/metrics/metric_api.rb
@@ -132,7 +132,7 @@ module Hawkular::Metrics
       # @return [Array[MetricDefinition]]
       def query(tags = nil)
         tags_filter = tags.nil? ? '' : "&tags=#{@client.tags_param(tags)}"
-        @client.http_get("/metrics/?type=#{@type}#{tags_filter}").map do |g|
+        @client.http_get("/metrics/?timestamps=true&type=#{@type}#{tags_filter}").map do |g|
           Hawkular::Metrics::MetricDefinition.new(g)
         end
       end


josejulio:
Some cassettes needs to be recorded again:

Could you please update those by doing:
```bash
VCR_UPDATE=1 bundle exec rspec spec/integration/metric_spec.rb -e 'Availability metrics Should update tags for Availability definition'
VCR_UPDATE=1 bundle exec rspec spec/integration/metric_spec.rb -e 'Gauge metrics Should update tags for gauge definition'
```

and to see that everything is fine:

```bash
bundle exec rspec spec/integration/metric_spec.rb
```

abonas:
should the client doc/readme be changed to reflect this?

pilhuhn:
@josejulio Wouldn't hand-editing the cassette be faster?
@abonas We should mention it. While he change should be opaque for the consumer of the gem, so it is not absolutely necessary. But we need to mention that older versions of the gem will misbehave going forward without that change. 

I wonder if that `timestamps=true` should be a default option on the method `def query(tags=nil, timestamps=true)` to directly allow for the future to mimic what metrics is doing by omitting the timestamps (explicitly)?

josejulio:
> pilhuhn commented a day ago
> @josejulio Wouldn't hand-editing the cassette be faster?

Yes, but might be prone to false positive errors. If possible we should refresh the whole cassette.

> I wonder if that timestamps=true should be a default option on the method def query(tags=nil, timestamps=true) to directly allow for the future to mimic what metrics is doing by omitting the timestamps (explicitly)?

That would be a good. The code will stay backwards compatible but having this new param in case we would like to omit the timestamps.
If @jsanda doesn't have the opportunity to do that I volunteer to do that.

jkremser:
https://github.com/hawkular/hawkular-client-ruby/pull/202 has been merged, so this can be now closed, right?

josejulio:
Right, Closing it.

Issue #101, opened 2016-07-04
Readme contains broken links to documentation


abonas:
the "more info" section contains several links to documentation, they seem to lead nowhere.
https://github.com/hawkular/hawkular-client-ruby#more-info

cc @josejulio

josejulio:
All the links are meant to be seen in the context of yardoc [1]. Check first link on the Readme[2]

I though it because of this other link[3] on the Changelog section[4]. Compare with the one in the documentation[2].

One way to avoid this (i think) would be to create a separate Readme.md for github.

[1] https://github.com/hawkular/hawkular-client-ruby#generate-documentation
[2] http://www.hawkular.org/hawkular-client-ruby/
[3] https://github.com/hawkular/hawkular-client-ruby/blob/master/file.CHANGES.html
[4] https://github.com/hawkular/hawkular-client-ruby#changelog

abonas:
link in item 3 above is 404..
and it should work on github as well since many users are reading the readme there.

josejulio:
Yes, item 3 doesn't work on github but works on the Hawkular documentation[1]. I didn't do that, I just followed that behavior. 
I'll check what can be done.

[1] http://www.hawkular.org/hawkular-client-ruby/

jkremser:
The links are meant to work from the context of the  [generated yardoc](http://www.hawkular.org/hawkular-client-ruby). Unfortunately, the rdoc on github doesn't support those github directives (`ifdef::env-github[]..`) that work in adoc and markdown so we can't simple have one version for 2 purposes. Also now the documentation has to be generated and pushed to the gh-pages branch manually after the gem release and for the previous version the docs were missing.

So, we can make it work on the README on GitHub in the first place and also have a simple bash/ruby script that changes those 3 links in the generated yard docs, this would be done in travis only during the release. I guess, the release build can be recognized by some environmental variable in Travis, i.e we are building a tag and the tag has some `x.y.z` format. Also the travis would need to push the result of the yardoc back to the `gh-pages` branch, so there must be also some github token used + some magic.

If there is an easier way to do that, please leave a comment. Otherwise, I'll do the ^

jkremser:
Also it'd be nice to have also older docs available on the website, not only the latest ones.

Something like:
* http://www.hawkular.org/hawkular-client-ruby/2.9.0/docs
* http://www.hawkular.org/hawkular-client-ruby/2.8.0/docs
[...]

PR #198, opened 2017-03-17
Automatic generation of documentation in Travis in case of release.

jkremser:
This PR basically does as described in these 2 comments:
* https://github.com/hawkular/hawkular-client-ruby/issues/101#issuecomment-286608438
* https://github.com/hawkular/hawkular-client-ruby/issues/101#issuecomment-286607493

There is also related PR to the gh-pages branch - https://github.com/hawkular/hawkular-client-ruby/pull/199

Also there is no need to have those travis specific scripts, docker-compose files, .coveralls configs and other files in the released gem, so I removed them from the white list here: https://github.com/hawkular/hawkular-client-ruby/pull/198/files#diff-a5831e69d9fde06f5c110544c9e14a1aR19

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ rvm:
 env:
   global:
     - ADMIN_TOKEN=4f4a1434-8cb3-11e6-ae22-56b6b6499611
+    # GH_TOKEN=...  used by hawkular-website-bot to publish the documentation
+    - secure: "EpOp9rU8GW2v/QyuDFRZQPQ2vGZ8vPsheVvLE7KSsu3UjNF8STiE50a2gmUW6KYQU+8jwSDHl8zUjIgpBko1tAUMd4NXNPMxZA+ybBLyUwQbbkTvFcG7FVFeB0UyjPQzOAw7qLm53HvknDv/P65OgkWXfjUVFWo4plgZKdwufDHyUqyN1ZBOUKu1w8HZ00HLnUWkhSiV9YG4RRFKJfEriElyv24L0RA9X8yVsS1eGr9MJhTzVgF0bAW6dSyxarZEo4qvLnqko98mva/wfGkmIBJBy7wxBSfy92flSh44s+s/nPzNKh1vZ9RF5lUekvh3Iw3pzZ9gb9wiNIvVM3Fa3jRl82fFUsGmpD2frE0cy4us+YodvPpAYn2EgGoVQnTJaFCTeMRbQpQ67j71JO6fBxGkIHb+TtdyieVajPGlRnJa6RDpT2QDiwuzhOOMvNkxO+6k+87ePUGu3geng7RXIBV3FqsjfkY79cDo3EsVg95BAa2rEuyqysbdVWox1VK0QuvBaFtNCbBoegnk0PX0uNMKmt1cLsxvz5zfwgAVPTEXsa8/FcDC2X6Ua+oVmSVZTEW1anEjQUjz+CYwWaBTwXknZvRNsM3x/kYskVyieDzehu71L5exZtg6+P+In6N/NXw7DL0SokPE4VbPng71eOlMW3cTDnnxeR1fmffABGg="
   matrix:
     - RUN_ON_LIVE_SERVER=1 \
       VCR_UPDATE=1 \
@@ -18,3 +20,5 @@ services:
   - docker
 before_install:
   - (test $RUN_ON_LIVE_SERVER = 1 && ./.travis/start_hawkular_services.sh -d && ./.travis/wait_for_services.rb) || echo "Skipping live server"
+after_success:
+  - ./.travis.docs.rb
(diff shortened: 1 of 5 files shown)


cfcosta:
LGTM

.travis.yml
```diff
@@ -7,6 +7,8 @@ rvm:
 env:
   global:
     - ADMIN_TOKEN=4f4a1434-8cb3-11e6-ae22-56b6b6499611
+    # GH_TOKEN=...  used by hawkular-website-bot to publish the documentation
+    - secure: "EpOp9rU8GW2v/QyuDFRZQPQ2vGZ8vPsheVvLE7KSsu3UjNF8STiE50a2gmUW6KYQU+8jwSDHl8zUjIgpBko1tAUMd4NXNPMxZA+ybBLyUwQbbkTvFcG7FVFeB0UyjPQzOAw7qLm53HvknDv/P65OgkWXfjUVFWo4plgZKdwufDHyUqyN1ZBOUKu1w8HZ00HLnUWkhSiV9YG4RRFKJfEriElyv24L0RA9X8yVsS1eGr9MJhTzVgF0bAW6dSyxarZEo4qvLnqko98mva/wfGkmIBJBy7wxBSfy92flSh44s+s/nPzNKh1vZ9RF5lUekvh3Iw3pzZ9gb9wiNIvVM3Fa3jRl82fFUsGmpD2frE0cy4us+YodvPpAYn2EgGoVQnTJaFCTeMRbQpQ67j71JO6fBxGkIHb+TtdyieVajPGlRnJa6RDpT2QDiwuzhOOMvNkxO+6k+87ePUGu3geng7RXIBV3FqsjfkY79cDo3EsVg95BAa2rEuyqysbdVWox1VK0QuvBaFtNCbBoegnk0PX0uNMKmt1cLsxvz5zfwgAVPTEXsa8/FcDC2X6Ua+oVmSVZTEW1anEjQUjz+CYwWaBTwXknZvRNsM3x/kYskVyieDzehu71L5exZtg6+P+In6N/NXw7DL0SokPE4VbPng71eOlMW3cTDnnxeR1fmffABGg="
```
abonas, line 11: hmmm are we ok with commiting publicly a security token?
@pilhuhn ?

josejulio: It seems to be encrypted using Travis API. I don't have the link to the docs handy but you can only use them on non-prs builds.

josejulio: https://docs.travis-ci.com/user/environment-variables/#Defining-encrypted-variables-in-.travis.yml

abonas:
> https://docs.travis-ci.com/user/environment-variables/#Defining-encrypted-variables-in-.travis.yml

it says there "Encrypted environment variables are not available to pull requests from forks due to the security risk of exposing such information to unknown code." - we are working with forks

jkremser:
> it says there "Encrypted environment variables are not available to pull requests from forks due to the security risk of exposing such information to unknown code." - we are working with forks

That's what we want. We don't want to allow any anonymous PR to push to our gh-pages branch. Just for sure there is additional check for that in the script itself [here](https://github.com/hawkular/hawkular-client-ruby/pull/198/files#diff-515e88bb27e03f935a08a01e14f1d448R7).

.travis.yml
jkremser: correct, it's encrypted by travis. Probably it's not super secure, because it depends on some secret key, but it's commonly used by our other projects.

.travis.yml
```diff
@@ -18,3 +20,5 @@ services:
   - docker
 before_install:
   - (test $RUN_ON_LIVE_SERVER = 1 && ./.travis/start_hawkular_services.sh -d && ./.travis/wait_for_services.rb) || echo "Skipping live server"
+after_success:
+  - ./.travis.docs.rb
```
josejulio, line 24: Wondering if we could use the [deploy](https://docs.travis-ci.com/user/customizing-the-build/#Deploying-your-Code) phase of travis.

We would have less troubles if adding more jobs/permutations to the [build matrix](https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix).

jkremser: I was looking into it but there were other issues, like very strong assumptions on what you want to do, which was basically uploading a file to somewhere.

I workaround it with a simple hack:
https://github.com/hawkular/hawkular-client-ruby/pull/198/files#diff-515e88bb27e03f935a08a01e14f1d448R17

I.e. it's being run only for the travis job that ends with '.4'

josejulio: > I workaround it with a simple hack:
> https://github.com/hawkular/hawkular-client-ruby/pull/198/files#diff-515e88bb27e03f935a08a01e14f1d448R17

That seems to be good.

I was suggesting that one because of this: https://docs.travis-ci.com/user/deployment/rubygems/
We might be able to also use before/after_deploy and have it as part of our deploy pipeline. 
e.g. build docs, post to rubygems, have a beer.

jkremser: I saw that doc but I would rather do the releases with Edgar deliberately, more than a side effect of a new tag. What if there is a tag that is not a release? @abonas what do you think?

abonas: I agree @Jiri-Kremser , I also find it more "careful" to manually release at the moment.

israel-hdez (approved):
LGTM

PR #202, opened 2017-03-23
[HAWKULAR-1206] Add timestamps parameter due to changes in HWKMETRICS…

josejulio:
…-625

- The new parameter is optional and defaults to true to avoid breaking any existing code.

Similar to #197 but with argument `options` which could have `:timestamps` and recorded the cassettes.

### DIFF
--- a/lib/hawkular/metrics/metric_api.rb
+++ b/lib/hawkular/metrics/metric_api.rb
@@ -129,10 +129,13 @@ module Hawkular::Metrics
 
       # Query metric definitions by tags
       # @param tags [Hash]
+      # @param options [Hash] Additional options to configure
+      # @option options [Boolean] :timestamps If timestamps should be included on the response. Defaults to true
       # @return [Array[MetricDefinition]]
-      def query(tags = nil)
+      def query(tags = nil, options = {})
+        timestamps = (options.key?(:timestamps) ? options[:timestamps] : true).to_s
         tags_filter = tags.nil? ? '' : "&tags=#{@client.tags_param(tags)}"
-        @client.http_get("/metrics/?type=#{@type}#{tags_filter}").map do |g|
+        @client.http_get("/metrics/?timestamps=#{timestamps}&type=#{@type}#{tags_filter}").map do |g|
           Hawkular::Metrics::MetricDefinition.new(g)
         end
       end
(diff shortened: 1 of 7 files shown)


lib/hawkular/metrics/metric_api.rb
```diff
@@ -129,10 +129,13 @@ def create(definition)
 
       # Query metric definitions by tags
       # @param tags [Hash]
+      # @param options [Hash] Additional options to configure
+      # @option options [Boolean] :timestamps If timestamps should be included on the response. Defaults to true
       # @return [Array[MetricDefinition]]
-      def query(tags = nil)
+      def query(tags = nil, options = {})
```
josejulio, line 135: @pilhuhn what do you think on having a `hash` instead of a `boolean` for other params that might be added in the future to configure the query?

lib/hawkular/metrics/metric_api.rb
```diff
@@ -129,10 +129,13 @@ def create(definition)
...
+        timestamps = options[:timestamps] || true
```
israel-hdez: With this you are allowing the callers to inject random data to the query string of the URL.
What I mean, if I call:
`query(nil, { timestamps: "random data" })`

The `timestamps` variable will have "`random data`" value; thus, URL built will be `metrics/?timestamps=random data&type=metric_type`

What is even more weird is that these calls are equivalent:
`query(nil, { timestamps: true })`
`query(nil, { timestamps: false })`

Because:
`timestamps = false || true` will evaluate to true (because it's an OR!!)
`timestamps = true || true` will evaluate to true (because there is no choice, ha)

and final URL will be `metrics/?timestamps=true&type=metric_type` for both cases and will return timestamps.
So... I think this line is not correct at all. May be you explicitly want to test against `true`:
`timestamps = !options.key?(:timestamps) || options[:timestamps] == true`

israel-hdez (changes requested):

lib/hawkular/metrics/metric_api.rb
josejulio: > israel-hdez 3 minutes ago Member
> With this you are allowing the callers to inject random data to the query string of the URL.
> What I mean, if I call:
> query(nil, { timestamps: "random data" })
> 
> The timestamps variable will have "random data" value; thus, URL built will be metrics/?timestamps=random data&type=metric_type

I don't think this is really a problem (injecting random data).

> 
> What is even more weird is that these calls are equivalent:
> query(nil, { timestamps: true })
> query(nil, { timestamps: false })
> 
> Because:
> timestamps = false || true will evaluate to true (because it's an OR!!)
> timestamps = true || true will evaluate to true (because there is no choice, ha)
> 

True, was trying to actually accept 'true' and 'false'.

> and final URL will be metrics/?timestamps=true&type=metric_type for both cases and will return timestamps.
> So... I think this line is not correct at all. May be you explicitly want to test against true:
> timestamps = options[:timestamps] == true

I was trying to support boolean strings ('true', 'false'). In case someone got a boolean string (db, rest request, etc) but not sure if that would ever happen.

josejulio: Just updated the code, please give another go at that.

israel-hdez (approved):
Looks good :ok_hand:

PR #204, opened 2017-04-05
Adding support to list alerts and events for multiple tenants

### LINKED ISSUE #176: Support for multi-tenant operations
There are going to be cross tenant apis for alerting e.g get alerts for all tenants. 
We would need to have some support for those in clients, currently created with one tenant in mind.

israel-hdez:
Supported endpoints are:
- hawkular/alerts/admin/alerts
- hawkular/alerts/admin/events

This fixes hawkular/hawkular-client-ruby#176

### DIFF
--- a/lib/hawkular/alerts/alerts_api.rb
+++ b/lib/hawkular/alerts/alerts_api.rb
@@ -246,8 +246,18 @@ module Hawkular::Alerts
     # @param [Hash]criteria optional query criteria
     # @return [Array<Alert>] List of alerts in the system. Can be empty
     def list_alerts(criteria = {})
+      alerts(criteria: criteria)
+    end
+
+    # List fired alerts
+    # @param [Hash] criteria optional query criteria
+    # @param [Array] tenants optional list of tenants. The elements of the array can be any object
+    #   convertible to a string
+    # @return [Array<Alert>] List of alerts in the system. Can be empty
+    def alerts(criteria: {}, tenants:nil)
       query = generate_query_params(criteria)
-      ret = http_get('/' + query)
+      uri = tenants ? '/admin/alerts/' : '/'
+      ret = http_get(uri + query, multi_tenants_header(tenants))
       val = []
       ret.each { |a| val.push(Alert.new(a)) }
       val
@@ -298,9 +308,26 @@ module Hawkular::Alerts
     #  thin        boolean, return lighter events (omits triggering data for trigger-generated events)
     # @param [Hash] criteria optional query criteria
     # @return [Array<Event>] List of events. Can be empty
-    def list_events(*criteria)
-      query = generate_query_params(*criteria)
-      http_get('/events' + query).map { |e| Event.new(e) }
+    def list_events(criteria = {})
+      events(criteria: criteria)
+    end
+
+    # List Events given optional criteria. Criteria keys are strings (not symbols):
+    #  startTime   numeric, milliseconds from epoch
+    #  endTime     numeric, milliseconds from epoch
+    #  eventIds    array of strings
+    #  triggerIds  array of strings
+    #  categories  array of strings
+    #  tags        array of strings, each tag of format 'name|value'. Specify '*' for value to match all values
+    #  thin        boolean, return lighter events (omits triggering data for trigger-generated events)
+    # @param [Hash] criteria optional query criteria
+    # @param [Array] tenants optional list of tenants. The elements of the array can be any object
+    #   convertible to a string
+    # @return [Array<Event>] List of events. Can be empty
+    def events(criteria: {}, tenants: nil)
+      query = generate_query_params(criteria)
+      uri = tenants ? '/admin/events' : '/events'
+      http_get(uri + query, multi_tenants_header(tenants)).map { |e| Event.new(e) }
     end
 
     # Inject an event into Hawkular-alerts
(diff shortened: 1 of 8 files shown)


josejulio:
@moolitayer is this what you need?
//cc @lucasponce

spec/integration/alerts_spec.rb
```diff
@@ -463,6 +475,17 @@ module Hawkular::Alerts::RSpec
...
 
+    it 'Should list alerts for multiple tenants' do
+      client = Hawkular::Alerts::Client.new(ALERTS_BASE, creds, options)
+
+      alerts = client.alerts(tenants: [:hawkular, :foo])
+
+      expect(alerts).to_not be_nil
+      expect(alerts.size).to be >= 4
+      expect(alerts.count { |a| a.tenantId == 'hawkular' }).to be >= 2
+      expect(alerts.count { |a| a.tenantId == 'foo' }).to be >= 2
```
josejulio: Can you put something like
```ruby
expect(alerts.count { |a| a.tenantId == 'hawkular' } + alerts.count { |a| a.tenantId == 'foo' }).to eq alerts.size
```
Just to be sure that we don't have alerts from any other tenant.

spec/integration/alerts_spec.rb
```diff
@@ -623,6 +655,17 @@ module Hawkular::Alerts::RSpec
...
 
+    it 'Should list events for multiple tenants' do
+      client = Hawkular::Alerts::Client.new(ALERTS_BASE, creds, options)
+
+      events = client.events(criteria: { 'thin' => true }, tenants: [:hawkular, :foo])
+
+      expect(events).to_not be_nil
+      expect(events.size).to be >= 24
+      expect(events.count { |e| e.tenantId == 'hawkular' }).to be >= 12
+      expect(events.count { |e| e.tenantId == 'foo' }).to be >= 12
```
josejulio: Can you put something like
```ruby
expect(events.count { |e| e.tenantId == 'hawkular' } + events.count { |e| e.tenantId == 'foo' }).to eq events.size
```
Just to be sure that we don't have events from any other tenant.

josejulio (approved):
LGTM
I'll leave to your consideration a couple comments for the tests.

spec/integration/alerts_spec.rb
israel-hdez: Instead of adding one more expectation I just changed `>=` by `eq`

spec/integration/alerts_spec.rb
israel-hdez: Hmm... Had to filter by `eventType`, because it is returning alerts. Don't know why, because alerts cleanup should already be made. Additionally, for `hawkular` tenant I couldn't change `>=` to `eq`. For hawkular tenant, some other events are being returned. May be because Hawkular is adding events to `hawkular` tenant to monitor itself?

josejulio: That's possible. 
Is also possible that someone running the tests already had some events on their system. 
Recording new cassettes under those conditions is annoying...  that's why i left the >= in other parts of the code.

josejulio:
This still looks good to me, if you don't have anything else to add I'll proceed to merge.

lib/hawkular/alerts/alerts_api.rb
```diff
@@ -246,8 +246,18 @@ def get_alerts_for_trigger(trigger_id) # TODO: add additional filters
     # @param [Hash]criteria optional query criteria
     # @return [Array<Alert>] List of alerts in the system. Can be empty
     def list_alerts(criteria = {})
```
josejulio, line 248: nitpicking, we could remove this method and just add an alias to alerts.

lib/hawkular/alerts/alerts_api.rb
```diff
@@ -298,9 +308,26 @@ def acknowledge_alert(alert_id, by = nil, comment = nil)
     #  thin        boolean, return lighter events (omits triggering data for trigger-generated events)
     # @param [Hash] criteria optional query criteria
     # @return [Array<Event>] List of events. Can be empty
-    def list_events(*criteria)
-      query = generate_query_params(*criteria)
-      http_get('/events' + query).map { |e| Event.new(e) }
+    def list_events(criteria = {})
```
josejulio, line 311: nitpicking, we could remove this method and just add an alias to events.

edit: I don't know if it could work with the `*criteria` syntax

israel-hdez:
@josejulio Nothing else :)

lib/hawkular/alerts/alerts_api.rb
israel-hdez: Not really, `list_alerts` is being called with a hash. In `alerts` method I used keyword arguments. In methods with keyword arguments, ruby tries to match hashes to keyword arguments. If you do the alias, `alerts_client.list_alerts('thin' => true)` (which is in the specs) will be equivalent to `alerts_client.alerts('thin' => true)` and ruby will complain with an exception saying that there is no 'thin' argument. You are forced to specify:
`alerts_client.alerts(criteria: {'thin' => true})`
So, signatures are not very compatible.

israel-hdez: May be we could deprecate `list_alerts` and `list_events`? I don't want to do that, because we would start a name-consistency war :sweat_smile: due all other methods.

josejulio: I guess is OK.

PR #205, opened 2017-04-18
WIP Replacing the websocket-client-simple gem 

rubenvp8510:


rubenvp8510:
This still fails on VCR tests because needs to update the vcr_websocket gem, @cfcosta Could you please start reviewing this? Thank you!!

lib/hawkular/websocket_client.rb
```diff
@@ -0,0 +1,82 @@
...
+    include EventEmitter
+    @channel = EventMachine::Channel.new
+    attr_accessor :url, :ws, :open, :options, :thread
+    attr_reader :open
+    alias_method :open?, :open
+    class << self
+      attr_accessor :channel
+    end
+
+    def self.running_evm
```
cfcosta: I don't like this method name, if it starts EventMachine, we should call it `#run` or something like that. We don't need to make any references to EventMachine, since it is also an implementation detail.

spec/integration/hawkular_client_spec.rb
```diff
@@ -295,8 +349,8 @@ module Hawkular::Client::RSpec
         end
       end
 
-      xit 'Should both reuse the websocket connection', :websocket do
-        WebSocketVCR.record(example, self) do
+      it 'Should both reuse the websocket connection', :websocket do
```
cfcosta: `it 'reuses the WebSocket connection'` would be a better name for this test. Also, it is not testing anything.

spec/integration/hawkular_client_spec.rb
```diff
@@ -258,13 +312,13 @@ module Hawkular::Client::RSpec
...
             on.failure do |error|
-              actual_data[:data] = {}
-              puts 'error callback was called, reason: ' + error.to_s
+              actual_data[:data] = error
             end
           end
 
           actual_data = wait_for actual_data
-          expect(actual_data['status']).to eq('OK')
+          expect(actual_data['status']).to eq('OK') unless @agent_immutable
```
cfcosta: I don't like this conditional on the test, immutable agents should be tested on a different test case.

spec/integration/hawkular_client_spec.rb
```diff
@@ -50,24 +60,28 @@ module Hawkular::Client::RSpec
       expect { @hawkular_client.tokens_lyst_feeds }.to raise_error(NoMethodError)
     end
 
-    context 'and URIs as input', vcr: { decode_compressed_response: true } do
+    context 'and URIs as input' do
       it 'Should work with URI' do
```
cfcosta: What exactly does this test? It might "work" with URI, but what exactly does this mean?

lib/hawkular/websocket_client.rb
rubenvp8510: OK, I'll rename it.

spec/integration/hawkular_client_spec.rb
rubenvp8510: My fault, I'll remove the commit that fixes the tests because I did another PR https://github.com/hawkular/hawkular-client-ruby/pull/208 for that,  put your tests comments on that PR please :), but without that fix, the hawkular client tests fails on this PR.

spec/integration/hawkular_client_spec.rb
rubenvp8510: I agree with you, we can remove this test.

spec/integration/hawkular_client_spec.rb
josejulio: I was trying (as i did that [agent_immutable code-style](https://github.com/hawkular/hawkular-client-ruby/pull/186) part on a separate PR) to have a single tests that could cover either cases. 

When running the live tests we only have immutable agents (as we use docker for testing). But when developing local we could either have an immutable agent or a non immutable one (when running outside of docker).

Any suggestion?

hawkularclient.gemspec
```diff
@@ -22,7 +22,9 @@ Gem::Specification.new do |gem|
   gem.require_paths = ['lib']
 
   gem.add_runtime_dependency('rest-client', '~> 2.0.0')
-  gem.add_runtime_dependency('websocket-client-simple', '~> 0.3.0')
+  gem.add_runtime_dependency('faye-websocket')
+  gem.add_runtime_dependency('eventmachine')
```
cfcosta, line 26: One thing we have to keep in mind is that eventmachine depends on native extensions, and the dependencies for it might not be available (or the server might be running on Thin, which has its own reactor), but its a minor issue.

cfcosta:
About the specs with conditionals, more tests is better, because since you're trying different permutations, you might as well have different failing tests if any of the cases is failing. The most important part of the tests is that they test only one thing, so even if one test is failing, the others still reflect what is being tested.

josejulio:
> About the specs with conditionals, more tests is better, because since you're trying different permutations, you might as well have different failing tests if any of the cases is failing. The most important part of the tests is that they test only one thing, so even if one test is failing, the others still reflect what is being tested.

I understand about the issue. But this is more dependent on the server than the client code. 
If the server we are testing against is answering something different than success but is immutable, our client tests should not fail. IMO the immutability code path should be really tested in the server code.

Maybe we can (on our live tests) expose that path by setting up a second hawkular server with the immutable flag set to false and run those tests. Not sure if that would be an overkill. 
Thoughts?

cfcosta:
I like the idea, but it doesn't fix the core issue: the class is doing too
much. We can extract an object from that to do that kind of behaviour so we
don't need to record different cassetes or do http requests every time we
run/change specs. IMHO, client code API should just be a facade for more
specialized objects that are not public-facing, but that's more of a
guideline than a rule.

Cheers,
Cainã Costa

israel-hdez:
@josejulio I mostly agree with @cfcosta. There should be one test for mutable and one for immutable. Currently, the tests are "adjusting" themselves to the given environment. For me, that is not desirable. The build in travis should also be able to test both cases; else, you are not testing everything. 
What I really don't like is that the tests are silently auto-adjusting to the environment. I'd rather to see a "pending" warning to not forget that there is something to improve on the tests.

josejulio:
I agree that the current scenario is not the desired one and I would appreciate if you could improve this part.

Maybe we should stick with one path on the tests and create new test suite for a mutable hawkular-services.
As for testing on travis, we could modify the testing script to download the latest hawkular-services distro (not docker) and run mutable tests on that one.

josejulio:
Anyway I suppose what we discussed here is outside of the scope of this PR.
`Replacing the websocket-client-simple gem`
thoughts @cfcosta @israel-hdez ?

israel-hdez:
@josejulio Yes, for me it is out of scope of this one.

jkremser:
@rubenvp8510 what needs to be done to make it work?

rubenvp8510:
@Jiri-Kremser  I just need to update the websocket-vcr gem. I start working on that, I'll continue and update this PR so we can merge if no one has anything comment to add.

jotak:
Just a remark here: I don't know if it's going do be merged, but if so, let's not forget to keep branch `hawkular-1259` updated as well.

PR #206, opened 2017-04-19
Fix broken links in index page listing versions of documentation

israel-hdez:


jkremser:
lgtm
thanks, mea culpa, I thought I fixed that

PR #207, opened 2017-04-21
Operation callbacks fix

israel-hdez:
Created branch 2.8.x on the repo from v2.8.0 tag. On top of that branch/tag, cherry-picked commit 49fc84a. I will need this released as v2.8.1.

This is required to fix BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1443005.

Note 1: v2.9.0 included that commit/fix, so there is no problem in that version.
Note 2: Travis is raising errors when testing on live servers. I don't know why that's happening :disappointed:. In travis log, something is echoed saying that hawkular-services was the one that failed to perform an operation. So, may be the problem is somewhere else?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ services:
   - docker
 before_install:
   - (test $RUN_ON_LIVE_SERVER = 1 && docker-compose up -d && ./.travis/wait_for_services.rb) || echo "Skipping live server"
+  - gem update bundler
 # Lets wait a bit for the agent to load everything it needs.
 before_script:
   - (test $RUN_ON_LIVE_SERVER = 1 && sleep 20s) || true
--- a/hawkularclient.gemspec
+++ b/hawkularclient.gemspec
@@ -34,6 +34,7 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency('rubocop', '= 0.34.2')
   gem.add_development_dependency('coveralls')
   gem.add_development_dependency('rack', '~> 1.6.4')
+  gem.add_development_dependency('pry-byebug')
 
   gem.rdoc_options << '--title' << gem.name <<
     '--main' << 'README.rdoc' << '--line-numbers' << '--inline-source'
--- a/lib/hawkular/operations/operations_api.rb
+++ b/lib/hawkular/operations/operations_api.rb
@@ -338,7 +338,8 @@ module Hawkular::Operations
         when "#{operation_name}Response"
           same_path = parsed[:data]['resourcePath'] == operation_payload[:resourcePath]
           # failed operations don't return the operation name from some strange reason
-          same_name = parsed[:data]['operationName'] == operation_payload[:operationName]
+          same_name = operation_payload[:operationName].nil? ||
+                      parsed[:data]['operationName'] == operation_payload[:operationName].to_s
           if same_path # using the resource path as a correlation id
             success = same_name && parsed[:data]['status'] == 'OK'
             success ? callback.perform(:success, parsed[:data]) : callback.perform(:failure, parsed[:data]['message'])


josejulio:
Do we really need to fix that with v2.8.1? Can't they use v2.9 ?

josejulio:
> Note 2: Travis is raising errors when testing on live servers. I don't know why that's happening . In travis log, something is echoed saying that hawkular-services was the one that failed to perform an operation. So, may be the problem is somewhere else?

Maybe the server (hawkular-services) is too new for v2.8. If we are going to support old versions we should consider on specifying a hawkular-services version to test against instead of using the latest version.  
see [docker-compose](https://github.com/hawkular/hawkular-client-ruby/blob/master/docker-compose.yml#L2)

israel-hdez:
@josejulio Well... I would say yes. But by doing that, some VCR cassettes in MiQ will need to be re-recorded. I want to avoid that since that BZ is happening in CFME 5.8.0, which is based in MiQ "fine", which is only receiving bug-fixes.
I will check if testing against an older server version fixes the tests.

pilhuhn:
@israel-hdez  " is echoed saying that hawkular-services was the one that failed to perform an operation."  I think the agent inside h-services is set to immutable (as we are running in-container in Travis) and thus operations can indeed not be peformed. I think @josejulio had a fix for this some time ago?

jkremser:
Supporting multiple minor versions and back-porting stuff may complicate the whole process, but sooner or later we  have to probably do that anyway.

@josejulio perhaps it make sense to run the live tests only for the master branch and PRs against master (?)

lgtm, I'd merge it despite the travis red, if anyone doesn't object

josejulio:
What versions of hawkular-services are we going to support with `hawkular-ruby-client` v2.8? If we are going to support latest version, then our live tests should pass on that version, else we won't be sure if all the tests run.

Live testing was working since v.2.8. If we aren't going to support latest (hawkular-services) version we should run the tests with the [latest hawkular-services](https://hub.docker.com/r/hawkular/hawkular-services/tags/) supported version by `hawkular-client-ruby` v2.8.

Having tests running on old cassettes (about 6 months ago) is almost as having no tests, but that's just my opinion :)

josejulio:
> @josejulio Well... I would say yes. But by doing that, some VCR cassettes in MiQ will need to be re-recorded. I want to avoid that since that BZ is happening in CFME 5.8.0, which is based in MiQ "fine", which is only receiving bug-fixes.

I'm not sure what would be the problem on recording the tests again, we might need to support less stuff this way.

israel-hdez:
> I'm not sure what would be the problem on recording the tests again, we might need to support less stuff this way.

@josejulio You resist to bring support, haha. But I understand you ;) I would have to sync with other people also using the gem, so that they also re-record their cassettes. I just want to avoid that hassle.

Anyway, tonight my pillow suggested me a way to fix the BZ in MiQ/fine side. I can apply that fix and we can forget about this PR. However, MiQ/fine will still be a "user" of v2.8.0 and that forces us to bring support to that version if something else arises.

josejulio:
> @josejulio You resist to bring support, haha. But I understand you ;) I would have to sync with other people also using the gem, so that they also re-record their cassettes. I just want to avoid that hassle.

I understand it can be a hassle. But think about it, If 2.8 is failing on live tests, its a hint that something may be wrong.
I'm just suggesting that if we aren't aiming to support latest hawkular-services with v2.8, we just set the right version on the docker-compose, so our tests will run fine (or fix the tests with https://github.com/hawkular/hawkular-client-ruby/pull/186 and maybe https://github.com/hawkular/hawkular-client-ruby/pull/191).
although I'm more in favor of bumping the version to 2.9, it should be backwards compatible since we are trying to use [semantic versioning](http://semver.org/).

israel-hdez:
@josejulio a short discussion about bumping the version happened in ManageIQ/manageiq-gems-pending#102. Even with semantic versioning the suggestion was to lock to a specific version, and that's the reason why I am keeping `fine` in 2.8.x.

Well.... Too much discussion about support/not-support previous versions. So, I placed ManageIQ/manageiq#14851 to fix the BZ in MiQ side.

We should agree on how to handle these kind of issues. If we are going to support previous versions, I can fix the specs as you @josejulio suggested in your last comment to have a green on 2.8.x branch.

josejulio:
> We should agree on how to handle these kind of issues. If we are going to support previous versions, I can fix the specs as you @josejulio suggested in your last comment to have a green on 2.8.x branch.

That would be awesome, thanks!

israel-hdez:
I'll close this. It doesn't apply anymore, because MiQ fine was updated to use 3.0.1.

PR #208, opened 2017-04-23
Fix hawkular client specs to take into account VCR_UPDATE.

rubenvp8510:
Also, take into account on some tests when the agent is immutable.


rubenvp8510:
@josejulio @cfcosta Could you please review?

spec/integration/hawkular_client_spec.rb
```diff
@@ -50,24 +60,28 @@ module Hawkular::Client::RSpec
...
 
       it 'Should work with URI on metrics client' do
         uri = URI.parse HOST
         opts = { tenant: 'hawkular' }
-        ::RSpec::Mocks.with_temporary_scope do
-          mock_metrics_version
-          the_client = Hawkular::Metrics::Client.new(uri, @creds, opts)
-          expect { the_client.http_get '/status' }.to_not raise_error
+        record('HawkularClient', nil, cassette_name) do
+          ::RSpec::Mocks.with_temporary_scope do
```
cfcosta: Why the temporary scope is needed here?

rubenvp8510: It's seems like is not necessary there, I just copy & paste

rubenvp8510:
Any other observations for this PR?

josejulio:
This looks good now.
@cfcosta If we are going to do something about that immutable_agent flag, i suggest to do in a separate PR and tackle all the places where is used.

WDYT?

jotak:
Do we plan to merge it soon? (I mean, like today :) )
It would fix some tests for #196 (and also has some conflicts with it)

rubenvp8510:
If no one has any more comments on this I think is probably that this could be merged soon (today) :D.

@josejulio @cfcosta there is any other comments?

josejulio:
LGTM

PR #210, opened 2017-04-25
Use hawkular tenant in hawkular_client_spec.rb

jotak:
Tests were failing when run with VCR_OFF=1 flag. Tenant should be given here.


rubenvp8510:
I did some general fixes to this tests on https://github.com/hawkular/hawkular-client-ruby/pull/208

PR #213, opened 2017-04-26
fixing typos in release notes and adding the release notes for 3.0.1

jkremser:
also removing the `link:` prefix from `README.rdoc` because it caused the url on http://www.hawkular.org/hawkular-client-ruby/docs/latest/ page to look like this:

`http://www.hawkular.org/www.hawkular.org/hawkular-client-ruby/docs/latest/file.api_breaking_changes.html`

which obviously doesn't work


josejulio:
LGTM

PR #215, opened 2017-04-26
Adds a timeout to wait_for_services script

josejulio:
To avoid 'waiting-for-services' to hang waiting.

Next step would be to use error code of waiting-for-services on hawkular-services build script to avoid failing if stuck on waiting-for-services


rubenvp8510:
LGTM

PR #216, opened 2017-05-02
Fix race condition while reading chunks

jotak:
Metrics reading could occur while writing chunks is not finished (ie. some chunks are written but not all), which would produce errors. We can detect that and return nil instead.

### DIFF
--- a/lib/hawkular/inventory/inventory_api.rb
+++ b/lib/hawkular/inventory/inventory_api.rb
@@ -445,26 +445,34 @@ module Hawkular::Inventory
     end
 
     def extract_structures_from_body(response_body_array)
-      response_body_array.map { |element| rebuild_from_chunks(element['data']) }
+      response_body_array.map { |element| Client.rebuild_from_chunks(element['data']) }
         .select { |full| full } # evict nil
-        .map { |full| decompress(full) }
+        .map { |full| Client.decompress(full) }
     end
 
-    def rebuild_from_chunks(data_node)
+    def self.rebuild_from_chunks(data_node)
       return if data_node.empty?
       master_data = data_node[0]
       return Base64.decode64(master_data['value']) unless (master_data.key? 'tags') &&
                                                           (master_data['tags'].key? 'chunks')
+      master_timestamp = master_data['timestamp']
       last_chunk = master_data['tags']['chunks'].to_i - 1
       all = Base64.decode64(master_data['value'])
       return if all.empty?
       (1..last_chunk).inject(all) do |full, chunk_id|
         slave_data = data_node[chunk_id]
+        # Race condition: slave_data might be nil if not all chunks have been written yet on DB
+        # Consider the object is not there yet
+        return nil unless slave_data
+        # The same race condition could give us an expirated slave_data; do sanity check to cover that
+        # Timestamps are useful here, they're in consecutive, decreasing order
+        slave_timestamp = slave_data['timestamp']
+        return nil unless slave_timestamp == master_timestamp - chunk_id
         full.concat(Base64.decode64(slave_data['value']))
       end
     end
 
-    def decompress(raw)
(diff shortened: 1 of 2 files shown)


jotak:
Note: this error has been seen once on the java agent integration tests (and PR sent there), but not (yet) on the ruby client, though it could happen here also in theory.

josejulio:
@jotak Rubocop is complaining about some code, could you please update it?

jotak:
oops, forgot to run it locally, sorry... will update

jotak:
@josejulio the test failures are then same that were mentioned by @pilhuhn yesterday, did you find out what's the cause?

josejulio:
Some tests were pointing directly to subsystem agent, latest version of hawkular services uses the hawkular javaagent. I updated the tests on #218

josejulio:
@jotak could you please rebase?

jotak:
done

josejulio:
Thanks!

josejulio:
Sorry for not asking until now, but could you put a (unit) test for it?

jotak:
@josejulio I'm adding a new file for unit tests (there was only integration test on that til now). Also changed rebuild_from_chunks to be a class method.

josejulio:
Awesome thanks!

jkremser:
@josejulio is this good to go?

lib/hawkular/inventory/inventory_api.rb
```diff
@@ -445,26 +445,34 @@ def extract_structure_from_body(response_body_array)
...
+        # The same race condition could give us an expirated slave_data; do sanity check to cover that
+        # Timestamps are useful here, they're in consecutive, decreasing order
+        slave_timestamp = slave_data['timestamp']
+        return nil unless slave_timestamp == master_timestamp - chunk_id
         full.concat(Base64.decode64(slave_data['value']))
       end
     end
 
-    def decompress(raw)
+    def self.decompress(raw)
```
jkremser, line 475: actually, couldn't we make it private? Same for `rebuild_from_chunks` or is it meant to be used outside of the client?

jotak: Just learnt about "private_class_method" :-)

jotak:
Hmm, still some work to do here...

jkremser:
@jotak perhaps it doesn't have to be a class method, but only normal instance method htat is private?

jotak:
I made it a class method because it has no dependency on the object's state, it looks cleaner to me but also easier to test with unit tests (the issue, if I have to instanciate the client in the unit test, is that the initializer performs http call to get the version... which sucks for a unit test)

Private class methods cannot be called from an instanciated object of that class, such as in java?

jkremser:
I see, I tried to make it instance private (https://pastebin.com/61bFZMG2) and the old tests passed, but you are right that initializing the client does the call.

Edit: perhaps it would worth adding some option in the hash that is passed to the client constructor that would omit the version check. But this is not related to your PR at all, I don't want to block it, on the other hand exposing internal method is no good either, because we would need to maintain them and preserve their backward-compatibility, even though they are only impl. detail

jotak:
That's probably a common case for java-ist doing ruby :) http://stackoverflow.com/questions/20674/is-there-a-way-to-call-a-private-class-method-from-an-instance-in-ruby

Well, if no objection, I'd rather go for making it public.

jkremser:
>  I'd rather go for making it public.

:+1: go ahead, I can take that in a follow up PR by adding option to the client that will not do the http check.

jotak:
Ok. But intuitively I would say it's a good practice to turn stateless methods into class methods, no? It's a way to formalize that they don't rely on the object state.

josejulio (approved):
This looks good to me. I don't have anything else to comment for this.

PR #217, opened 2017-05-02
WIP - Show trace that we are ignoring mutable or immutable tests

josejulio:


josejulio:
@israel-hdez @cfcosta 
Could you please comment on this WIP ?
This is meant to 

1. Fetch the agent (either hawkular wildfly agent or javaagent). Hawkular-services is now using javaagent, so I'm updating to fetch the agent (the old path and the type are irrelevant to the tests, we just want the feed agent)
2. Explicitly do a set of tests that run on mutable agent or immutable agent. Since both tests are the same but the output is different. I wrapped the actual test in a function and calling them depending on the context.

spec/integration/operations_spec.rb
```diff
@@ -169,34 +176,51 @@ module Hawkular::Operations::RSpec
...
-          expect(actual_data['status']).to eq('OK') unless @agent_immutable
-          expect(actual_data['message']).to start_with('Added JDBC Driver') unless @agent_immutable
-          expect(actual_data['driverName']).to eq(driver_name) unless @agent_immutable
-          expect(actual_data).to include('Command not allowed because the agent is immutable') if @agent_immutable
+          wait_for actual_data
+        end
+
+        context 'Mutable agent' do
+          before(:each) do
+            skip 'mutable agent specs don\'t run on immutable agent' if @agent_immutable
```
cfcosta, line 205: I still don't like this. We should duplicate the setup to create one mutable and one immutable, if that's what it takes. Tests should be predictable, and any behaviour that introduces variants should be extracted into different tests/different contexts.

spec/integration/operations_spec.rb
```diff
@@ -134,9 +134,20 @@ module Hawkular::Operations::RSpec
...
 
       describe 'Operation/Operation' do
-        before(:all) do
-          @random_uuid = SecureRandom.uuid
-          @not_so_random_uuid = 'not_so_random_uuid'
+        let(:random_uuid) do
+          SecureRandom.uuid
+        end
+
+        let (:not_so_random_uuid) do
```
cfcosta, line 141: Why do we need this? `let` declarations are memoized thorough the test case, so while it is random, it is consistent.

cfcosta (changes requested):

spec/integration/operations_spec.rb
josejulio: I suppose that's what it needs to be done. run two servers, group the tests and run mutable tests on a second server / feed.

spec/integration/operations_spec.rb
josejulio: I need a random uuid, one for the whole run is OK.

spec/spec_helper.rb
```diff
@@ -196,17 +196,22 @@ def host(type)
...
+      resource_type_id: 'Hawkular WildFly Agent'
+    )
+    inventory.list_resources_for_type(resource_type_path, true).first
   end
 
-  def in_container(inventory, resource_path)
-    properties = inventory.get_resource(resource_path, true).properties
+  def agent_in_container?(agent)
+    properties = agent.properties
     properties.key?('In Container') ? (properties['In Container'] == 'true') : false
```
cfcosta, line 210: `properties['In Container'] == 'true'`

spec/spec_helper.rb
```diff
@@ -196,17 +196,22 @@ def host(type)
...
   end
 
-  def immutable(inventory, resource_path)
-    properties = inventory.get_resource(resource_path, true).properties
+  def agent_immutable?(agent)
+    properties = agent.properties
     properties.key?('Immutable') ? (properties['Immutable'] == 'true') : false
```
cfcosta, line 215: Same here.

spec/spec_helper.rb
cfcosta: Or even better, `agent.properties['In Container'] == 'true'`, and we don't even need the temporary variable.

spec/integration/operations_spec.rb
israel-hdez: I think @cfcosta  is saying that you can remove `not_so_random_uuid`. Instead, use `random_uuid` everywhere and the tests will be fine.

josejulio:
Closing as we may follow other path.

israel-hdez (commented):
@cfcosta has already pointed everything.
Is there any documentation for hawkular operations API? I couldn't find anyting in hawkular's website.

spec/integration/operations_spec.rb
josejulio: Actually i need `not_so_random_uuid` to stay the same between runs. I suppose the name could be changed to `fixed_id` or something like that.

josejulio:
@israel-hdez  some info is on the following links:

https://github.com/hawkular/hawkular-commons/tree/master/hawkular-command-gateway/hawkular-command-gateway-api
https://github.com/hawkular/hawkular-commons/tree/master/hawkular-command-gateway/hawkular-command-gateway-api/src/main/resources/schema

PR #218, opened 2017-05-02
Updating specs to pass on latest hawkular-services with hawkular java…

josejulio:
…agent

The scope of this PR is to make the tests compatible when running on against a hawkular-services with javaagent [0.37.0.Final](https://github.com/hawkular/hawkular-services/releases/tag/0.37.0.Final).

The agent is on a different path. 
The API to fetch the resources types is now used to successfully select any of the two agent types.
Some other tests required small changes.

// cc @pilhuhn


josejulio:
Finally green...

Had to modify the hawkular-javaagent-config.yaml from the `hawkular-services` image a bit to scan more frequently.

josejulio:
Also, this should make [hawkular-bot](https://github.com/hawkular/hawkular-services/pull/154#issuecomment-298724097) happy.

josejulio:
@israel-hdez could you take a look please?

spec/integration/inventory_spec.rb
```diff
@@ -152,7 +152,7 @@ module Hawkular::Inventory::RSpec
         it 'Should list resources for feed' do
           resources = @client.list_resources_for_feed feed_id
 
-          expect(resources.size).to be(2)
+          expect(resources.size).to be >= 2
```
israel-hdez, line 155: Hope some day we can find a way to make this more reliable.

.travis/wait_for_services.rb
```diff
@@ -40,6 +40,6 @@
     end
   end
 end
-puts 'Waiting 2 minutes for agent to complete it\'s first round...'
-sleep 120
+puts 'Waiting 1 minute for agent to complete it\'s first round...'
+sleep 60
```
israel-hdez, line 44: Is there a way to find if the agent already did its first round?
I don't like too much the idea of just waiting and then hope that it is done.

israel-hdez (approved):
It looks good.
I just placed a some questions that you may know how to answer.

.travis/wait_for_services.rb
josejulio: Not really sure, 1 minute should be more than enough on these tests.
I guess there is not a really way to know when the agent has finished (except for reading that every resource is in place, which is what the tests does).
Anyway most of the tests have a wait mechanism to not fail immediately but give some time before actually failing.

spec/integration/inventory_spec.rb
josejulio: Well, we must ask ourselves what are we really testing here, the fact that we should get feeds or that hawkular-services should have 2 (or 4, as the current configuration of the javaagent does).

If both, we can configure our hawkular-services's agent (as [done](https://github.com/josejulio/hawkular-client-ruby/blob/b4d5450fedd13ea055bc2a5f00ab752bed5d57ff/.travis/build_config_for_testing.rb) in this PR to increase its scan frequency) to be more specific on what we expect of it. 

IMO that would be more fit on a separate PR on which we should decide what we really want to test and how to configure it.

israel-hdez: Yes, that's for another PR.

Some days ago, I was thinking that we should manually place metrics, alerts and inventory under another tenant and don't rely on the agent. Using another tenant will make the test data "safe" from being updated because of the agent running in hawkular services.

Whatever we are trying to test here, it is related to "feeds". And feeds is a general concept in inventory, not agent specific. We should be able to manually register a feed and turn this test into a predictable one.

israel-hdez:
I'm merging if nothing else is needed.

spec/integration/inventory_spec.rb
josejulio: Just take in considerations that the same tests also run on every `hawkular-services` PR  ( @pilhuhn asked me to do that  ) as a way to early check for changes that would break compatibility with the client.

As of inventory.3 we are not longer able to write data into inventory (yet), the only capable of that are the agents.

We can make these tests predictable by updating hawkular-agent-configuration.yaml

PR #219, opened 2017-05-03
Escaping feed_id before using it in list_metric_types

israel-hdez:
list_metric_types assumes that the received feed_id is unescaped and
escapes it to form a CanonicalPath. Because the passed feed_id comes
from a CanonicalPath, it is already escaped and must be unescaped to
avoid double unescaping it.

NOTE: This should be released as 3.0.2. It is causing an error in MiQ.


israel-hdez:
@Jiri-Kremser can you take a look and release, if you are ok?

jkremser:
lgtm

PR #220, opened 2017-05-04
Release notes and readme

jkremser:
I fixed also some links in the readme file, because some are not valid because of the renaming of the clients and some are not working from this context http://www.hawkular.org/hawkular-client-ruby/docs/latest/

@israel-hdez could you pls merge?


israel-hdez:
Looks good

PR #222, opened 2017-06-07
Error name format unification

cfcosta:
Changes from `Hawkular::BaseClient::HawkularException` format to `Hawkular::Exception`, and also changes all errors thrown to be instances of `Hawkular::Exception`, so client code can rescue just one exception and capture all errors thrown by the gem.

### DIFF
--- a/CHANGES.rdoc
+++ b/CHANGES.rdoc
@@ -4,6 +4,10 @@ This document describes the relevant changes between releases of the
 _hawkular-client_ project.
 
 
+=== v 3.0.3
+* Standardized Exceptions under the Hawkular namespace, and the old names were deprecated, here is the list:
+1. <code>Hawkular::BaseClient::HawkularException</code> -> <code>Hawkular::Exception</code>
+2. <code>Hawkular::BaseClient::HawkularConnectionException</code> -> <code>Hawkular::ConnectionException</code>
 === v 3.0.2
 * Adding <code>URI.unescape</code> in the <code>list_metrics_for_resource</code> method to fix the behavior of the method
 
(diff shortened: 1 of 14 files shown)


CHANGES.rdoc
```diff
@@ -4,6 +4,10 @@ This document describes the relevant changes between releases of the
 _hawkular-client_ project.
 
 
+=== v 3.0.3
+* Standardized Exceptions under the Hawkular namespace, and the old names were deprecated, here is the list:
+1. <code>Hawkular::BaseClient::HawkularException</code> -> <code>Hawkular::Exception</code>
+2. <code>Hawkular::BaseClient::HawkularConnectionException</code> -> <code>Hawkular::ConnectionException</code>
```
jkremser: looks good, but I think it is a breaking change from the API point of view so this will probably require the major version bump

cfcosta: I don't think it is a breaking change because the API does not change, as the old exceptions are still available. Maybe a minor version bump instead of a patch one?

josejulio: Its a thin line. 
Before we were raising RuntimeError exceptions and now we are raising Hawkular::ArgumentError.

Still, If a tree falls in a forest and no one is around to hear it, does it make a sound?
(as in, did someone used this? Should we raise the major version?)

[1] https://github.com/hawkular/hawkular-client-ruby/pull/222/files#diff-27ed7a1113ec9bf87df507d3990b581eR40

abonas: the "signature" of the api changes because now another type of exception is returned, so it can be definitely considered as a breaking change.
I don't know if anyone uses it though.
@pilhuhn any idea?

cfcosta:
Closing and reopening, Ruby 2.2 segfaulted.

PR #223, opened 2017-06-08
Connect operations client lazily

cfcosta:
This pull request fixes a 500 page that happens when you try to do a power
operation when the middleware server is offline, by making the connection be
created lazily when an operation is called.

The idea behind this is to allow operations code to catch errors when connecting to the server, because while errors from the operations can be catched.

As an example, here's this piece of code:

```ruby
connection.operations(true).undeploy(deployment_data) do |on|
  on.success do |data|
    _log.debug "Success on websocket-operation #{data}"
    emit_middleware_notification(:mw_op_success, 'Undeploy', deployment_name, ems_ref, MiddlewareDeployment)
  end
  on.failure do |error|
    _log.error 'error callback was called, reason: ' + error.to_s
    emit_middleware_notification(:mw_op_failure, 'Undeploy', deployment_name, ems_ref, MiddlewareDeployment)
  end
end
```

If the error happens on `#undeploy`, then the failure clause will be called, but if the error happens on `#operations(true)`, which spawns a new connection, it will be raised as normal.

As a bonus, this PR also removes the sleep code from the connection handler,
which makes both the tests and real world usage much, much, much faster.

Before:

        hawkular-client-ruby git:(fix-operations-500-error) $ time RUBYOPT="-Ilib -W0 -r./bundle/bundler/setup.rb" rspec spec/integration/operations_spec.rb
        [Coveralls] Set up the SimpleCov formatter.
        [Coveralls] Using SimpleCov's default settings.
        Run options: exclude {:skip=>true}
        ...................

        Finished in 32.57 seconds (files took 0.69792 seconds to load)
        19 examples, 0 failures

After:

        hawkular-client-ruby git:(fix-operations-500-error) ✗ $ time RUBYOPT="-Ilib -W0 -r./bundle/bundler/setup.rb" rspec spec/integration/operations_spec.rb
        [Coveralls] Set up the SimpleCov formatter.
        [Coveralls] Using SimpleCov's default settings.
        Run options: exclude {:skip=>true}
        ...................

        Finished in 3.34 seconds (files took 0.74947 seconds to load)
        19 examples, 0 failures

Bugzilla Ticket: https://bugzilla.redhat.com/show_bug.cgi?id=1444940


cfcosta:
@Jiri-Kremser @abonas can you take a look at this?

lib/hawkular/operations/operations_api.rb
```diff
@@ -79,22 +83,34 @@ def initialize(args)
...
+    end
 
-      creds = args[:credentials]
-      base64_creds = ["#{creds[:username]}:#{creds[:password]}"].pack('m').delete("\r\n")
+    def connect
+      return if @connecting || (@ws && @ws.open?)
+
+      @connecting = true
+
+      condition = new_cond
```
abonas: where the `new_cond` comes from and what does it represent?

lib/hawkular/operations/operations_api.rb
```diff
@@ -335,7 +354,6 @@ def invoke_operation_helper(operation_payload, operation_name = nil, binary_cont
     end
 
     def check_pre_conditions(hash = {}, params = [], &callback)
-      fail 'Handshake with server has not been done.' unless @ws.open?
```
abonas, line 338: Why is this statement removed? I'm pretty sure there's a reason but following the changes in the PR doesn't point on it very clearly to me...

lib/hawkular/operations/operations_api.rb
```diff
@@ -107,12 +123,13 @@ def initialize(args)
         end
       end
 
-      sleep args[:wait_time]
+      Timeout.timeout(@wait_time) { condition.wait_while { !@ws.open? } }
+      @connecting = false
```
abonas: if something fails between the part where `connecting` is set to `true`, to the part where `connecting` set to `false`, then won't the `connecting` remain `true` when it shouldn't?

lib/hawkular/operations/operations_api.rb
cfcosta: Because the client tries to connect again before doing any operation, so we don't need to check if we are connected or not.

lib/hawkular/operations/operations_api.rb
cfcosta: That's true!

lib/hawkular/operations/operations_api.rb
cfcosta: It comes from `MonitorMixin`, http://ruby-doc.org/stdlib-2.4.1/libdoc/monitor/rdoc/MonitorMixin.html#method-i-new_cond , and generates a Mutex that is used later with `wait_until` to keep it locked until we finish connecting.

abonas:
@israel-hdez please have a look

lib/hawkular/operations/operations_api.rb
abonas: so shouldn't this be handled somehow? :)

cfcosta: It is fixed, this patch is outdated :D

lib/hawkular/operations/operations_api.rb
```diff
@@ -322,6 +339,8 @@ def update_collection_intervals(hash, &callback)
     private
 
     def invoke_operation_helper(operation_payload, operation_name = nil, binary_content = nil, &callback)
+      synchronize { connect }
```
israel-hdez, line 334: I don't know where was the source of the exception. So, I'm guessing...

With your changes, I would expect `connect` to not raise an error even if the server rejects the connection or is down. And, then, I would expect the "connection refused" exception to be handled by `@ws.send` called in the later lines by calling the failure callback. I cannot figure any other way on how the callback may be invoked in case of any error. 

If my guessings are right, it's all ok. But as @mtho11 suggested in the other PR, it would be great if you can add a test to ensure that the code won't raise the "connection refused" exception, but invoke the failure callback. Specially, because @rubenvp8510 is doing work in replacing the websocket gem.

israel-hdez (commented):
Looks fine, I have only one comment.

lib/hawkular/operations/operations_api.rb
cfcosta: That's true, actually, and it is something that was not covered by tests, so I fixed it.

cfcosta:
@israel-hdez @abonas did some changes, can you take a look at it again?

abonas:
@pilhuhn please have a look

pilhuhn:
I think that makes sense. I recall that we battled with getting the error callback right in the past for cases when credentials are wrong and similar.

israel-hdez:
@cfcosta I have no more comments. Ready to merge?

PR #225, opened 2017-06-28
Upgrade C* version 3.0.12

jotak:
As per h*services doc, it's the current recommended version

cf http://www.hawkular.org/hawkular-services/docs/installation-guide/#_apache_cassandra


cfcosta:
LGTM.

PR #226, opened 2017-07-13
Fix issue with mixed JSON and binary data in Hawkular websocket message

israel-hdez:
When invoking "export JDR" operation, Hawkular sends through the
websocket the required JSON response for any operation. Along with the
JSON response, the binary data of the report is also sent. As a
consequence, when parsing the JSON response also the binary data was
being parsed causing an exception.

These changes determine if there is binary data in the response from
Hawkular to parse only the JSON part and separate the binary data
fragment.


lib/hawkular/operations/operations_api.rb
```diff
@@ -33,9 +33,31 @@ class Client < Hawkular::BaseClient
     # helper for parsing the "OperationName=json_payload" messages
     class WebSocket::Frame::Data
       def to_msg_hash
-        chunks = split('=', 2)
+        operation_name, json = split('=', 2)
```
josejulio, line 36: +1

lib/hawkular/operations/operations_api.rb
```diff
@@ -33,9 +33,31 @@ class Client < Hawkular::BaseClient
...
 
-        { operationName: chunks[0], data: JSON.parse(chunks[1]) }
+        # Check if there is a zip file following JSON.
+        # This check is done only in the first 100KB, hoping it's unlikely to
+        # have such large amount of JSON before a zip file is attached.
+        magic_bits = ["\x50\x4B\x03\x04", "\x50\x4B\x05\x06", "\x50\x4B\x07\x08"]
+        search_chunk = json[0, 102_400]
```
josejulio: According to [this code](https://github.com/josejulio/hawkular-agent/blob/f3b9238af581cea8b5c2f3fe5b7dab41ed58a581/hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/cmd/FeedCommProcessor.java#L289), when binary data is present, the `content-type` gets changed to `application/vnd.okhttp.websocket+binary`.

Maybe we could search only when binary data is know to be present?

lib/hawkular/operations/operations_api.rb
```diff
@@ -105,7 +127,8 @@ def connect
         client.on(:message, once: true) do |msg|
           parsed_message = msg.data.to_msg_hash
 
-          logger.log("Sent WebSocket message: #{parsed_message}")
+          # Not working with websocket-vcr
```
josejulio: What's not working?

lib/hawkular/operations/operations_api.rb
```diff
@@ -33,9 +33,31 @@ class Client < Hawkular::BaseClient
...
     class WebSocket::Frame::Data
       def to_msg_hash
-        chunks = split('=', 2)
+        operation_name, json = split('=', 2)
 
-        { operationName: chunks[0], data: JSON.parse(chunks[1]) }
+        # Check if there is a zip file following JSON.
+        # This check is done only in the first 100KB, hoping it's unlikely to
+        # have such large amount of JSON before a zip file is attached.
+        magic_bits = ["\x50\x4B\x03\x04", "\x50\x4B\x05\x06", "\x50\x4B\x07\x08"]
```
josejulio: Can you put a commentary about what those `magic_bits` are? 
I'm supposing they are magic bits for some compressing formats, maybe would be worth to specify which ones.

lib/hawkular/operations/operations_api.rb
```diff
@@ -33,9 +33,31 @@ class Client < Hawkular::BaseClient
     # helper for parsing the "OperationName=json_payload" messages
     class WebSocket::Frame::Data
       def to_msg_hash
-        chunks = split('=', 2)
+        operation_name, json = split('=', 2)
 
-        { operationName: chunks[0], data: JSON.parse(chunks[1]) }
+        # Check if there is a zip file following JSON.
+        # This check is done only in the first 100KB, hoping it's unlikely to
```
josejulio: Sounds reasonable, I don't think there is other way to know this. Unless there is a library that is able to parse only one json object from a string and ignore the rest.

josejulio (commented):
Looks good! I only have some comments/questions.

lib/hawkular/operations/operations_api.rb
israel-hdez: Due to [the way the callback is invoked](https://github.com/shokai/event_emitter/blob/e7ad10c0755c0427d69c3653caac4c2a222adce4/lib/event_emitter/emitter.rb#L53), `logger` becomes out of scope and undefined.

josejulio: Isn't there any other way to log the message?

israel-hdez: I guess it can be done [like in the handle_message method](https://github.com/israel-hdez/hawkular-client-ruby/blob/48aeda8b1beacf91b7bd2fea64968403fd0b1755/lib/hawkular/operations/operations_api.rb#L399) where a new instance of logger is created.

BTW, I just realized that [the `@session_id`](https://github.com/israel-hdez/hawkular-client-ruby/blob/48aeda8b1beacf91b7bd2fea64968403fd0b1755/lib/hawkular/operations/operations_api.rb#L135) is not being set because is also out of scope. It's not used anywhere in the gem. So, not sure if we should worry about it. If we want to fix it, the better way should be to place a PR on event_emitter hoping that the author is willing to change that.

lib/hawkular/operations/operations_api.rb
israel-hdez: Yes! It can be done :)

israel-hdez: Hmm.... It can be done only in "live". It fails with VCR because it's mocking the `message` and there is no way to retrieve if it's binary or text.
So, I think I'll leave it as is for now.

lib/hawkular/operations/operations_api.rb
israel-hdez: Comments are now in place. All are for ZIP files. The first one is what should be always expected, at least with JDR report operation.

lib/hawkular/operations/operations_api.rb
josejulio: Maybe we should remove that session_id altogether (if could be used in the future we could fill an issue to seek alternatives in a separate PR).
Was this broken before or is a consequence of something new?

josejulio: BTW, can we do the new instance of logger? Logging that message could be useful...

israel-hdez: Ok, logging is back, and session_id is removed.
Looks like session_id has been broken for some time. But we didn't realized that something was wrong. Notice that logging is before setting the session_id. The out-of-scope variable was causing an exception but ruby throws it silently because the exception happens in a secondary thread. So, indeed, the line setting the session_id was never being reached.

josejulio (approved):

josejulio:
LGTM

PR #227, opened 2017-07-14
Upgrading the simple-websocket-vcr from 0.0.7. to 0.1.0

jkremser:
@israel-hdez could you pls look? This is needed for the JDR reports, right?


israel-hdez:
LGTM
Yes, it's needed. Small change, I'll merge :)

PR #228, opened 2017-07-28
Add "delete_immediately" parameter to export_jdr

israel-hdez:
This specifies to delete the copy of the JDR that is created in the
server node in a temporary directory once the report is received.


lib/hawkular/operations/operations_api.rb
```diff
@@ -326,12 +326,15 @@ def add_jdbc_driver(hash, &callback)
     # Exports the JDR report
     #
     # @param [String] resource_path canonical path of the WildFly server
+    # @param [Boolean] delete_from_server specifies whether the temporary file at the remote server should be deleted.
```
abonas: I'm wondering if it's necessary to call it delete_from_server. Why simply "delete" is not enough? maybe I missed some additional use case?

lib/hawkular/operations/operations_api.rb
```diff
@@ -326,12 +326,15 @@ def add_jdbc_driver(hash, &callback)
...
     # @param callback [Block] callback that is run after the operation is done
-    def export_jdr(resource_path, &callback)
+    def export_jdr(resource_path, delete_from_server = false, &callback)
       fail Hawkular::ArgumentError, 'resource_path must be specified' if resource_path.nil?
       check_pre_conditions(&callback)
 
-      invoke_specific_operation({ resourcePath: resource_path }, 'ExportJdr', &callback)
+      invoke_specific_operation({ resourcePath: resource_path,
+                                  deleteImmediately: delete_from_server },
```
abonas: same here, it reads a bit strange to me `deleteImmediately = delete_from_server` (not disputing the fact it works, just the code readability aspect) . it just sounds like there's another option "delete_from_somewhere_else"

cfcosta: Agree, `#delete_immediately` would make more sense.

cfcosta (changes requested):
LGTM, but I'd change the `delete_from_server` param to another, more descriptive name.

lib/hawkular/operations/operations_api.rb
israel-hdez: Ok, changed the name to delete_immediately.

abonas:
tests are failing

josejulio:
This require a hawkular-services release with latest agent.

israel-hdez:
@josejulio Yes, I realized that yesterday when I was trying to record with docker builds. I will ask if it can be released. At least, recorded scenario now works.

cfcosta (approved):

docker-compose.yml
```diff
@@ -1,5 +1,5 @@
 hawkular:
-  image: "hawkular/hawkular-services:latest"
+  image: "hawkular/hawkular-services:devel"
```
josejulio: Not sure of this:
Using devel means that we will testing against master hawkular-services, thus if there is problem (on services), it will make our build hard fail.
On the other hand it will fail earlier, as opposed to a services release, but Hawkular-services PRs already run a test round against master version of the gem (see [here](https://github.com/hawkular/hawkular-services/pull/171#issuecomment-319419439)).
I was also thinking that we could lock a hawkular-services version for test (i.e., version instead of latest), and bumping as we support newer versions.

israel-hdez: Right now, it's the only way to make tests pass since there isn't a release with latest hawkular-commons code.
But, I'm also not sure about using devel. I both like the idea and dislike it :speech_balloon:

josejulio:
Is already green, do you plan to do something else or is good to merge?

israel-hdez:
Nothing else. It's good to go.

PR #232, opened 2017-10-04
First step for new inventory api

josejulio:


josejulio:
@jotak Is feasible to get resources that have certain metric_id?
Consider `fetch_availabilities_for` [1] that fetches the availabilities by metric_id, then it does some canonical path magic [2] to get the resource.

In this scenario, It would be good if we could get all the resources that have certain metric_id on their list.

If that's not possible, I suppose i could list the resources by resource_type and manually test if they have the availability. This would be a bit harder. i.e. `Server Availability` can be found on Wildfly Servers and Domain Wildfly Servers, more could be added later. Other option would be to use the provided servers list [3] in this case to get the Resource types and then check for the metric.

[1] https://github.com/ManageIQ/manageiq-providers-hawkular/blob/master/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb#L190
[2] https://github.com/ManageIQ/manageiq-providers-hawkular/blob/master/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb#L208
[3] https://github.com/ManageIQ/manageiq-providers-hawkular/blob/master/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb#L191

// cc: @jshaughn @lucasponce

jmazzitelli:
I don't know the direct answers to these but let me say the following:

1) the new inventory has no concept of "canonical paths" - its all just opaque resource IDs
2) the new inventory maps resources to metrics but there is no notion of any special metric of type "availability"

"Availability" metrics is something we need to think about - how, for example, will we model this via Prometheus metrics?

lucasponce:
> In this scenario, It would be good if we could get all the resources that have certain metric_id on their list.

I think that perhaps we should use some kind of tags mechanism at resource level and add metrics names there to be indexed.

Is this case only for availabilities or it is more generic ?

If it is only for availability, a simpler solution is to have a hasAvailability at resource level, then the indexed flag is just a boolean.

Prometheus metrics I think is a following discussion as in the timeline we are still using hMetrics, if we mixed the discussion perhaps we can add complexity (but I think is worth to think on it).

Note that in hServices, backfill services are back with the new inventory.

jotak:
@josejulio , I looked at this case previously and found that at some point, there is some hardcoding pointing out items that provide an availability metrics: it's through the `AVAIL_TYPE_ID` constant defined in classes, here [1] and there [2].
So, I'm thinking if these classes could hold in the same way their resource type name so we can know that resource type XX has avail metric YY. For instance, in `middleware_server.rb` we would have `TYPE_ID = 'Whatever Server'`. But I don't know if one class refers to one resource type, or if there could be several of them (and we wouldn't want to hardcode it, then).

That way, when looking for avail metrics, instead of querying for all resources that have avail metrics, we query for all resources that have type containing avail metrics (and we get these metrics in Resource::type::metrics).

[1] https://github.com/ManageIQ/manageiq-providers-hawkular/blob/master/app/models/manageiq/providers/hawkular/middleware_manager/middleware_deployment.rb#L3
[2] https://github.com/ManageIQ/manageiq-providers-hawkular/blob/master/app/models/manageiq/providers/hawkular/middleware_manager/middleware_server.rb#L3

If that doesn't work, maybe Lucas' proposal of tagging system is the right way.

jshaughn:
Perhaps we should have a short meeting to talk about the actual use cases of the provider's use of availability.  Remember that going forward with prometheus we don't plan on having any avail metrics at all, only possibly leveraging the UP timeseries provided for the instance (i.e. server).  This could be a useful checkpoint of avail now and going forward.

lib/hawkular/inventory/entities_v4.rb
```diff
@@ -0,0 +1,95 @@
+# It contains class definitions that are used by the inventory REST client
+module Hawkular::InventoryV4
```
israel-hdez, line 2: Perhaps we could baptise the new inventory as `Hawkular::Registry`. I think _registry_ is more appropiate :)

jotak:
I suggest we merge this PR so that we get the starting point, and have a meeting today, as suggested by Jay to solve the issues raised ?

jshaughn:
Merging this checkpoint.  The WIP is not yet expected to pass CI.

PR #234, opened 2017-10-11
Minimal tests with inventory v4

jotak:
First step to validate the api, but for the time being it doesn't use cassettes, security context etc.
Hard-coded server url

### DIFF
--- a/lib/hawkular/inventory/inventory_api_v4.rb
+++ b/lib/hawkular/inventory/inventory_api_v4.rb
@@ -1,5 +1,5 @@
 require 'hawkular/base_client'
-# require 'json'
+require 'uri'
 
 require 'hawkular/inventory/entities_v4'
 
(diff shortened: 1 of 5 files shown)


jotak:
Note: the tests are passing for me, but I need to start my own hawkular server build aside (no hawkular-services docker image built yet)

jotak:
PS: I just built a docker image for hawkular-services with new inventory: `jotak/hawkular-services:0.40.0.Final-SNAPSHOT` ... cc @josejulio

lib/hawkular/inventory/inventory_api_v4.rb
```diff
@@ -37,29 +37,29 @@ def self.create(hash)
     # Get single resource by id
     # @return Resource the resource
     def resource(id)
-      hash = http_get("/resources/#{id}")
+      hash = http_get("/resources/#{URI.escape(id)}")
```
cfcosta, line 40: I think http_get should escape the URI directly, instead of adding it on each call.

cfcosta (changes requested):
Looks good, but I think changing the URI escaping to be more consistent is necessary.

lib/hawkular/inventory/inventory_api_v4.rb
jotak: that would be a big refactor, as `http_get` is used in other subclasses and some of them perform a non-standard escaping that couldn't be combined with URI.escape (see https://github.com/hawkular/hawkular-client-ruby/blob/master/lib/hawkular/client_utils.rb#L20 )

lib/hawkular/inventory/inventory_api_v4.rb
```diff
@@ -37,29 +37,29 @@ def self.create(hash)
...
-      hash = http_get("/resources/#{id}/tree")
+      hash = http_get("/resources/#{URI.escape(id)}/tree")
       Resource.new(hash)
     end
 
     # List root resources
     # @return [Array<Resource>] List of resources
     def root_resources
-      # FIXME: pagination
+      # FIXME: pagination => lazy-loaded list with ruby?
```
josejulio, line 54: +1

lib/hawkular/inventory/inventory_api_v4.rb
josejulio: Agree, lets review that on a separate PR

PR #235, opened 2017-10-11
Use ERB::Util.url_encode which also escapes '/'

josejulio:
'/' is used on id's and isn't escaped by URI.escape.
 Maybe there is others that aren't handled and is the reason
 `client_utils.rb` was born.

We can't do the escaping on `http_get` unless we pass around the `params`, but we would need to replicate that to `http_post`, `http_put`, etc.

### DIFF
--- a/spec/unit/base_spec.rb
+++ b/spec/unit/base_spec.rb
@@ -255,3 +255,32 @@ describe 'Base Spec' do
     end
   end
 end
+
+describe 'Url builder' do
+  it 'Should encode simple url' do
+    expect(Hawkular::BaseClient.new('not-needed').url('resource/endpoint')).to eql('resource/endpoint')
+  end
+
+  it 'Should encode with one param' do
+    expect(Hawkular::BaseClient.new('not-needed').url('resource/%s', 5)).to eql('resource/5')
+  end
+
+  it 'Should work with booleans' do
+    expect(Hawkular::BaseClient.new('not-needed').url('resource/%s', true)).to eql('resource/true')
+  end
+
+  it 'Should work with symbols' do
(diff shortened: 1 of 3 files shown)


josejulio:
@cfcosta Is safe to use ERB on the gem?

spec/unit/base_spec.rb
```diff
@@ -255,3 +255,27 @@
...
+    expect(Hawkular::BaseClient.new('not-needed').url('resource/%s', true)).to eql('resource/true')
+  end
+
+  it 'Should work with symbols' do
+    expect(Hawkular::BaseClient.new('not-needed').url('abc/%s', :true)).to eql('abc/true')
+  end
+
+  it 'Should work with multiple params' do
+    expect(Hawkular::BaseClient.new('not-needed').url('resource/%s/inner/%s', true, :ok)).to eql('resource/true/inner/ok')
+  end
```
jotak: You could add a test to make also sure it passes with "&" and "?" for query params

jotak:
If this encoding ignores "?" and "&", it LGTM

josejulio:
> If this encoding ignores "?" and "&", it LGTM

It doesn't, but it only encodes the params passed. e.g.

`url('resource/%s/stuff?myid=%s&this=that', 'super', '0&delete_everything=true')` will only encode `super` and `0&delete_everything=true`resulting in:
`resource/super/stuff?myid=0%26delete_everything%3Dtrue&this=that`

spec/unit/base_spec.rb
josejulio: Done

jotak:
Oh ok, I see
Looks good to me.

cfcosta (approved):

PR #236, opened 2017-10-11
Generic resources endpoint

josejulio:
Depends on  #235

### DIFF
--- a/lib/hawkular/inventory/inventory_api_v4.rb
+++ b/lib/hawkular/inventory/inventory_api_v4.rb
@@ -57,14 +57,26 @@ module Hawkular::InventoryV4
     # @return [Array<Resource>] List of resources
     def root_resources
       # FIXME: pagination => lazy-loaded list with ruby?
-      http_get('/resources?root=true')['results'].map { |r| Resource.new(r) }
+      resources root: "true"
+    end
+
+    # List resources
+    # @param [Hash] filter options to filter the resource list
+    # @option filter :root If truthy, only get root resources
+    # @option filter :feedId Filter by feed id
+    # @option filter :typeId Filter by type id
+    def resources(filter = {})
+      # FIXME: pagination => lazy-loaded list with ruby?
+      filter[:root] = !! filter[:root] if filter.key? :root
+      filter_query = '?' + filter.keys.join('=%s&') + '=%s' unless filter.empty?
+      http_get(url("/resources#{filter_query}", *filter.values))['results'].map { |r| Resource.new(r) }
     end
 
     # List resources for type
     # @return [Array<Resource>] List of resources
     def resources_for_type(type)
       # FIXME: pagination => lazy-loaded list with ruby?
-      http_get(url('/resources?typeId=', type))['results'].map { |r| Resource.new(r) }
+      resources typeId: type
     end
 
     # Return version and status information for the used version of Hawkular-Inventory


lib/hawkular/inventory/inventory_api_v4.rb
```diff
@@ -37,35 +36,47 @@ def self.create(hash)
...
+    # List resources
+    # @param [Hash] filter options to filter the resource list
+    # @option filter :root If truthy, only get root resources
+    # @option filter :feedId Filter by feed id
+    # @option filter :typeId Filter by type id
+    def resources(filter = {})
+      # FIXME: pagination => lazy-loaded list with ruby?
+      filter[:root] = !! filter[:root] if filter.key? :root
+      filter_query = '?' + filter.keys.join('=%s&') + '=%s' unless filter.empty?
+      http_get(url("/resources#{filter_query}", *filter.values))['results'].map { |r| Resource.new(r) }
```
josejulio, line 72: @jotak can you review this part?

lib/hawkular/inventory/inventory_api_v4.rb
```diff
@@ -37,35 +36,47 @@ def self.create(hash)
...
     end
 
     # List resources for type
     # @return [Array<Resource>] List of resources
     def resources_for_type(type)
       # FIXME: pagination => lazy-loaded list with ruby?
-      http_get("/resources?typeId=#{URI.escape(type)}")['results'].map { |r| Resource.new(r) }
+      resources typeId: "type"
```
jotak: `type`, not `"type"` (ie. not a string), I guess

lib/hawkular/inventory/inventory_api_v4.rb
jotak: I've no strong opinion about opening the API to queries with feedId, however note that I didn't put a "get_by_feedid" function in a first place, because I think MiQ should be able to perform all of its queries without having to deal with the feedId. (Except for operations, that's why it's still in the model)

According to Lucas, the feedId leaking in MiQ code was more of a technical debt due to early versions. If we can run queries without it, we would mechanically reduce the number of calls.

jotak: (but the ruby client is not just for MiQ so I think it's fine)

lib/hawkular/inventory/inventory_api_v4.rb
josejulio: Whoops

lib/hawkular/inventory/inventory_api_v4.rb
```diff
@@ -57,14 +57,26 @@ def children_resources(parent_id)
...
+    end
+
+    # List resources
+    # @param [Hash] filter options to filter the resource list
+    # @option filter :root If truthy, only get root resources
+    # @option filter :feedId Filter by feed id
+    # @option filter :typeId Filter by type id
+    def resources(filter = {})
+      # FIXME: pagination => lazy-loaded list with ruby?
+      filter[:root] = !! filter[:root] if filter.key? :root
```
cfcosta, line 70: `filter.fetch(:root) { false }`

lib/hawkular/inventory/inventory_api_v4.rb
```diff
@@ -57,14 +57,26 @@ def children_resources(parent_id)
...
+      filter_query = '?' + filter.keys.join('=%s&') + '=%s' unless filter.empty?
+      http_get(url("/resources#{filter_query}", *filter.values))['results'].map { |r| Resource.new(r) }
     end
 
     # List resources for type
     # @return [Array<Resource>] List of resources
     def resources_for_type(type)
       # FIXME: pagination => lazy-loaded list with ruby?
-      http_get(url('/resources?typeId=', type))['results'].map { |r| Resource.new(r) }
+      resources typeId: type
```
cfcosta, line 79: I don't like this, really. It makes for the params to be un-ruby. If we're going to use the params in any way, maybe they should be changed to be `type_id`, and be translated on the client layer.

josejulio: I was going to use https://apidock.com/rails/String/camelize but not sure if possible.

lib/hawkular/inventory/inventory_api_v4.rb
josejulio: I need to cast it to boolean also, any suggestion?

lib/hawkular/inventory/inventory_api_v4.rb
cfcosta: Not yet, we had some talks on including ActiveSupport exactly for that, but I think for the time being we can just copy the method implementation.

lib/hawkular/inventory/inventory_api_v4.rb
cfcosta: `!!`, if you really really need a boolean

PR #237, opened 2017-10-14
Adds children by type InventoryV4::Resource method

josejulio:
Useful when having the whole tree and want to get certain types

### DIFF
--- a/lib/hawkular/inventory/entities_v4.rb
+++ b/lib/hawkular/inventory/entities_v4.rb
@@ -82,6 +82,16 @@ module Hawkular::InventoryV4
       @_hash = hash.dup
     end
 
+    def children_by_type(type, recursive = false)
+      fail Hawkular::ArgumentError 'Resource tree not loaded, load it by calling resource_tree' if @children.nil?
+      selected_children = []
+      @children.each do |child|
+        selected_children << child if child.type.id == type
+        selected_children.push(*child.children_by_type(type, recursive)) if recursive
+      end
+      selected_children
+    end
+
     def ==(other)
       self.equal?(other) || other.class == self.class && other.id == @id
     end
--- a/lib/hawkular/inventory/entities_v4.rb
+++ b/lib/hawkular/inventory/entities_v4.rb
@@ -82,14 +82,16 @@ module Hawkular::InventoryV4
       @_hash = hash.dup
     end
 
-    def children_by_type(type, recursive = false)
+    def children(recursive = false)
+      return @children unless recursive == true
       fail Hawkular::ArgumentError 'Resource tree not loaded, load it by calling resource_tree' if @children.nil?
-      selected_children = []
-      @children.each do |child|
-        selected_children << child if child.type.id == type
-        selected_children.push(*child.children_by_type(type, recursive)) if recursive
+      @children.flat_map do |child|
+        [child, *child.children(recursive)]
       end
-      selected_children
+    end
+
+    def children_by_type(type, recursive = false)
+      children(recursive).select { |c| c.type.id == type }
     end
 
     def ==(other)
--- a/spec/unit/inventory_v4_spec.rb
+++ b/spec/unit/inventory_v4_spec.rb
@@ -20,25 +20,35 @@ describe 'Inventory v4' do
     }
   end
 
-  describe 'children by type' do
-    it 'should return direct children' do
-      r = Resource.new(resource_hash)
-      expect(r.children_by_type('type-02').size).to eq(1)
+  let(:resource) do
+    Resource.new(resource_hash)
+  end
+
+  describe '#children' do
+    it 'returns direct children' do
+      expect(resource.children.size).to eq(3)
     end
 
-    it 'should return 0 when direct children type is not found' do
-      r = Resource.new(resource_hash)
-      expect(r.children_by_type('type-03').size).to eq(0)
+    it 'returns all children' do
+      expect(resource.children(true).size).to eq(6)
+    end
+  end
+
+  describe '#children_by_type' do
+    it 'returns direct children' do
+      expect(resource.children_by_type('type-02').size).to eq(1)
     end
 
-    it 'should work recursive' do
-      r = Resource.new(resource_hash)
-      expect(r.children_by_type('type-02', true).size).to eq(3)
+    it 'returns 0 when direct children type is not found' do
+      expect(resource.children_by_type('type-03').size).to eq(0)
     end
 
-    it 'should work recursive when the type is not on top' do
-      r = Resource.new(resource_hash)
-      expect(r.children_by_type('type-03', true).size).to eq(1)
+    it 'works recursive' do
+      expect(resource.children_by_type('type-02', true).size).to eq(3)
+    end
+
+    it 'works recursive and does not matter if the type is not on top' do
+      expect(resource.children_by_type('type-03', true).size).to eq(1)
     end
   end
 end
--- a/spec/unit/inventory_v4_spec.rb
+++ b/spec/unit/inventory_v4_spec.rb
@@ -1,0 +1,44 @@
+require_relative '../spec_helper'
+
+include Hawkular::InventoryV4
+describe 'Inventory v4' do
+  let(:resource_hash) do
+    {
+      'id' => 'root',
+      'type' => {},
+      'children' => [
+        { 'id' => 'child-01', 'type' => { 'id' => 'type-01' }, 'children' => [] },
+        { 'id' => 'child-02', 'type' => { 'id' => 'type-02' },
+          'children' => [
+            { 'id' => 'child-03', 'type' => { 'id' => 'type-02' }, 'children' => [] },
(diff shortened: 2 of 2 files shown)


josejulio:
@jotak could you please review?

spec/unit/inventory_v4_spec.rb
```diff
@@ -0,0 +1,47 @@
+require_relative '../spec_helper'
+
+module Hawkular::Inventory::RSpecUnit
```
cfcosta: Why is this module necessary?

spec/unit/inventory_v4_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      end
+
+      it 'should return 0 when direct children type is not found' do
+        r = Resource.new(resource_hash)
+        expect(r.children_by_type('type-03').size).to eql(0)
+      end
+
+      it 'should work recursive' do
+        r = Resource.new(resource_hash)
+        expect(r.children_by_type('type-02', recursive=true).size).to eql(3)
```
cfcosta: Named arguments in Ruby do not work like this. Also, by convention we use `eq` instead of `eql`, because `eql` means equality by identity, not equality by comparison.

spec/unit/inventory_v4_spec.rb
josejulio: Not really sure :) I saw other inventory specs do the same.
If you think is safe to be removed, then I can do that.

spec/unit/inventory_v4_spec.rb
```diff
@@ -0,0 +1,47 @@
+require_relative '../spec_helper'
+
+module Hawkular::Inventory::RSpecUnit
+  include Hawkular::InventoryV4
+  context 'Inventory v4 unit test' do
```
cfcosta: If this is supposed to be added to other test files, then it should be using [shared examples](https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples). If that's not the case, a block describing tests should be `describe`, not `context`. And no need to name it `X unit test`.

cfcosta (changes requested):

spec/unit/inventory_v4_spec.rb
cfcosta: If it is not, then it should be changed before moving forward. Even if that's correct for some reason, specs should not be contained in modules because that would make the setup different from the usage because of namespacing.

lib/hawkular/inventory/entities_v4.rb
```diff
@@ -82,6 +82,15 @@ def initialize(hash)
       @_hash = hash.dup
     end
 
+    def children_by_type(type, recursive = false)
+      selected_children = []
+      @children.each do |child|
+        selected_children << child if child.type.id == type
+        selected_children.push(*child.children_by_type(type, recursive)) if recursive
+      end unless @children.nil?
```
jotak: Maybe the call should fail when children is nil, because it doesn't mean there's no children, it means they haven't been loaded.

josejulio: Makes sense to me.

spec/unit/inventory_v4_spec.rb
```diff
@@ -0,0 +1,45 @@
+require_relative '../spec_helper'
+
+include Hawkular::InventoryV4
+describe 'Inventory v4' do
+
```
cfcosta: No need for this blank line.

spec/unit/inventory_v4_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+          ]
+        },
+        { "id" => "child-06", "type" => { "id" => "type-01"} }
+      ]
+    }
+  end
+
+  describe 'children by type' do
+    it 'should return direct children' do
+      r = Resource.new(resource_hash)
```
cfcosta: This should be extracted to a `let`.

lib/hawkular/inventory/entities_v4.rb
```diff
@@ -82,6 +82,15 @@ def initialize(hash)
       @_hash = hash.dup
     end
 
+    def children_by_type(type, recursive = false)
+      selected_children = []
+      @children.each do |child|
+        selected_children << child if child.type.id == type
+        selected_children.push(*child.children_by_type(type, recursive)) if recursive
```
cfcosta: No way to do it iteractively with a `flat_map`? Something like this might work:

```ruby
selected_children = @children.select { |x| x.type.id == type }

if recursive
  selected_children.flat_map! { |x| x.children_by_type(type, recursive) }
end

selected_children
```

cfcosta (changes requested):
Getting there!

lib/hawkular/inventory/entities_v4.rb
```diff
@@ -82,6 +82,16 @@ def initialize(hash)
...
     end
 
+    def children_by_type(type, recursive = false)
+      fail Hawkular::ArgumentError 'Resource tree not loaded, load it by calling resource_tree' if @children.nil?
+      selected_children = []
+      @children.each do |child|
+        selected_children << child if child.type.id == type
+        selected_children.push(*child.children_by_type(type, recursive)) if recursive
+      end
+      selected_children
```
cfcosta: No way to do it iteractively with a `flat_map`? Something like this might work:

```ruby
selected_children = @children.select { |x| x.type.id == type }

if recursive
  selected_children.flat_map! { |x| x.children_by_type(type, recursive) }
end

selected_children
```

spec/unit/inventory_v4_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+            { 'id' => 'child-04', 'type' => { 'id' => 'type-02' }, 'children' => [] },
+            { 'id' => 'child-05', 'type' => { 'id' => 'type-03' }, 'children' => [] }
+          ]
+        },
+        { 'id' => 'child-06', 'type' => { 'id' => 'type-01' }, 'children' => [] }
+      ]
+    }
+  end
+
+  describe 'children by type' do
```
cfcosta: `describe '#children_by_type'`, since we're testing a method.

jshaughn: @josejulio This is useful if you do already have the whole tree, but will you always have the whole tree in advance?  Would it make sense to have underlying support in the inventory UI to fetch resources by type?

josejulio: I can already fetch resources by type in the API. But in order to link them together, I have to group them by e.g. the feed_id. 

I'm trying to avoid searching by feed, instead fetching a whole tree and processing. (discarding the tree after I finished the processing it)
That way I'm sure that the whole tree belongs to the same "feed".

Then extracting the required types from the tree (depending what is currently being processed).

lib/hawkular/inventory/entities_v4.rb
josejulio: Will check, but i don't think that would work.

`selected_children = @children.select { |x| x.type.id == type }`
This will select the children with the required `type`, this is good.

`selected_children.flat_map! { |x| x.children_by_type(type, recursive) }`
This will only iterate over the selected children, this is bad, I want to check in all the children.

josejulio:
```ruby
selected_children = @children.select { |x| x.type.id == type }
if recursive
  selected_children.push(*@children.flat_map { |x| x.children_by_type(type, recursive) })
end
```
Is this better?
Although i preferred to do that in a single loop, but if this is more rubyist then is ok for me.
@cfcosta

josejulio:
@cfcosta updated the code, please take a look again.

spec/unit/inventory_v4_spec.rb
jotak: Don't hesitate to suggest API improvement on hawkular side. For instance, given our current indexing, I think we could perfectly add a "get child resource by type" query, given a type and a parent id. I'm not sure if it's more (or less) interesting than the "get tree" that is being filtered client-side (depends if you also need that tree information for later / other types)

lib/hawkular/inventory/entities_v4.rb
```diff
@@ -82,6 +82,18 @@ def initialize(hash)
       @_hash = hash.dup
     end
 
+    def children(recursive = false)
+      return @children unless recursive == true
+      fail Hawkular::ArgumentError 'Resource tree not loaded, load it by calling resource_tree' if @children.nil?
```
cfcosta, line 87: I think this is a syntax error, no? If not, ignore this comment.

spec/unit/inventory_v4_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      ]
+    }
+  end
+
+  let(:resource) do
+    Resource.new(resource_hash)
+  end
+
+  describe '#children' do
+    it 'should return direct children' do
```
cfcosta: No action necessary on this right now, but we’re planning on standardize test names, per recommended on betterspecs.org. In this case, it’d be something like “returns direct children”.

cfcosta (approved):
LGTM, there are some small suggestions but I think it’s good enough if you don’t want to change anything anymore.

PR #238, opened 2017-10-16
Inventory model changes

jotak:
- Remove Resource.children_ids
- Add Resource.config
- Add helper methods Metric.hawkular_id and hawkular_type
- Fix a test

### DIFF
--- a/lib/hawkular/inventory/entities_v4.rb
+++ b/lib/hawkular/inventory/entities_v4.rb
@@ -13,6 +13,14 @@ module Hawkular::InventoryV4
       @type = hash['type']
       @properties = hash['properties'] || {}
     end
+
+    def hawkular_id
+      @properties['hawkular.metric.id']
+    end
+
+    def hawkular_type
+      @properties['hawkular.metric.type']
+    end
   end
 
   class Operation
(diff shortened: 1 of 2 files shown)


jotak:
@abonas , @josejulio : there's some little changes in inventory model.
- children_ids are not available anymore because we've changed the way resources are stored internally, but you can still get the full resource tree when querying on resources/{id}/tree

- I've added back the "config" map from the previous model ; I thought that 'properties' and 'config' could be 1 map, but after discussion it was decided to make them separated again. I think MiQ only needs one of them, but I'm not sure which one...

Is it ok for you?

josejulio:
What's the difference between config and properties? what's the intended usage?

jotak:
@josejulio "config", mazz confirmed

jotak:
actually, if "properties" is useless in MiQ (it's used internally in hawkular), maybe we can hide it from the ruby client API

josejulio:
I still wonder what kind of values properties might contain :)

lib/hawkular/inventory/entities_v4.rb
```diff
@@ -13,6 +13,14 @@ def initialize(hash)
       @type = hash['type']
       @properties = hash['properties'] || {}
     end
+
+    def hawkular_id
```
cfcosta: Should `hawkular_id` and `hawkular_type` be always present? If so it's better to use `fetch` instead of `[]`.

jotak: They should be always present... for now. Not later when prometheus comes in the game.

cfcosta: Is prometheus related to inventory as well? Thought it was supposed to be contained in Metrics only. Either way I don't think it is a good idea to return nil on this.

josejulio: This is a "step" that i suppose we are going to remove when prometheus migration is done.
Each resource has a list of metrics, and this tells the "metric_id" where this metric is stored.

e.g. Resource A and B have both a x_count, but A.x_count is stored in `hawk_{a.id}_{x.id}.{hash}` and b in `hawk_{a.id}_{x.id}.{hash}`.

cfcosta (approved):
LGTM

PR #239, opened 2017-10-17
hawkular-1259: remove old inventory module, start updating itests

jotak:


jotak:
WIP - I started to fix itests but there's still a long way down. So, I let people choose if it's better to merge fast (in order to minimize git conflicts due to inventory renaming) or keep it here until it's 100% done

josejulio:
If i understand it correctly, this PR removes the old inventory and replaces it with inventory v4. Updating some tests on the way.

+1 for a quick merge (on hawkular-1259)

jotak:
So I had some time to commit more :)
Remarks on the current state:
- operation_api should be compliant now with new inventory...

- ... however I had to change some behaviour that will impact callers: previously, operations on deployments were expecting caller to provide the deployment resource id (whereas CMDGW expects server resource id), and the ruby client did the transformation [deployment=>server] thanks to CanonicalPath.up. This is no longer possible, so the caller will have to provide the server ID directly. @josejulio , can you check if that can be a problem?

- I fixed several rubocop issues but 3 are still there, including some weird complains about alignment where I can't see the problem (or my eyes are not aligned :) )

- Some tests are still failing:
  - `HawkularClient Should err on bad credentials` => that's because authentication is not yet activated in the docker image, it's not a real problem
  - <strike>`HawkularClient and Operations client Should both work the same way` => I don't know what's wring here. Got a java-side "Illegal argument exception" while sending a custom operation "Redeploy". Need to investigate more and/or get help here.</strike> FIXED (Redeploy was recently removed)
  - `Add JDBC driver should add the driver` (while running in docker with flag immutable=false) I think it's a known issue, and it disappears when I switch on immutability.
  - <strike>`Remove datasource should be performed and eventually respond with success` => Here we have a (small) design problem, due to resource IDs being now opaque. The test must be rewritten, because we don't know the ID of the entity that we want to remove. My suggestion is that we add the newly created resource ID in the CMDGW response (that is run through another test), and store this ID client-side.
  - `Remove JDBC driver should be performed and eventually respond with success` => same as above</strike> FIXED
  - Other failing tests are token related, I think we don't really care... (I've always seen them failing)

jotak:
Oh, and I forgot: the `Export JDR` test does not work. Need to investigate, too.

lib/hawkular/base_client.rb
```diff
@@ -42,7 +42,7 @@ def initialize(entrypoint = nil,
     end
 
     def url(url_format, *params)
-      url_format % params.map { |p| ERB::Util.url_encode(p) }
+      url_format % params.map { |p| hawk_escape(p) }
```
josejulio: Did this fail for some cases?

jotak: oops, I didn't mean to commit that... good catch!

lib/hawkular/operations/operations_api.rb
```diff
@@ -222,18 +221,16 @@ def undeploy(hash, &callback)
...
     #
     # @param callback [Block] callback that is run after the operation is done
     def enable_deployment(hash, &callback)
-      required = [:resource_path, :deployment_name]
+      required = [:resource_id, :feed_id, :deployment_name]
       check_pre_conditions hash, required, &callback
 
-      cp = ::Hawkular::Inventory::CanonicalPath.parse hash[:resource_path]
-      server_path = cp.up.to_s
-      hash[:resource_path] = server_path
```
josejulio, line 236: @jotak I suppose this is what you meant on the comments.

It confuses me that even the documentation stated that it required the Wildfly server (e.g. the documentation is the same now, or not?)

jotak: I believed that it changed several times in the past and the doc wasn't updated... Not sure, I didn't check git history...

josejulio: Thanks, I'll check how is handled on MIQ

cfcosta:
>So, I let people choose if it's better to merge fast (in order to minimize git conflicts due to inventory renaming) or keep it here until it's 100% done

@jotak I think it is better to keep it here until it is 100% done. Also, to move some resources to do proper reviewing.

josejulio:
> ... however I had to change some behaviour that will impact callers: previously, operations on deployments were expecting caller to provide the deployment resource id (whereas CMDGW expects server resource id), and the ruby client did the transformation [deployment=>server] thanks to CanonicalPath.up. This is no longer possible, so the caller will have to provide the server ID directly. @josejulio , can you check if that can be a problem?

I don't suppose it will be a big issue, need to update the code, yes. Will defer this `up` operations to MIQ code. (By using the db, not parsing anything)

josejulio:
> Other failing tests are token related, I think we don't really care... (I've always seen them failing)

AFAIK token is no longer used, we could remove them.

jshaughn:
> I don't suppose it will be a big issue, need to update the code, yes. Will defer this up operations to MIQ code. (By using the db, not parsing anything)

You could also get the parent by querying inventory for the deployment resource, which has the parentId.  If we can simplify the client code by adding support in the gem or in the backend, it can be good.

jotak:
I think parentId is not exposed on API at the moment, it's more an implementation detail for building the tree, but yeah it could be exposed.

lib/hawkular/operations/operations_api.rb
josejulio: I suppose we can receive the deployment_id and query the inventory for the server id?
That would ideally happen on hawkular-services if possible.
WDYT @jotak ?

josejulio:
> So, I let people choose if it's better to merge fast (in order to minimize git conflicts due to inventory renaming) or keep it here until it's 100% done
> @jotak I think it is better to keep it here until it is 100% done. Also, to move some resources to do proper reviewing.

I need to update some of this code to keep the refactoring of https://github.com/ManageIQ/manageiq-providers-hawkular/pull/75 which is blocking hawkular-1259

Can we please do a review round on this? 
Note that this is not going to be merged on master, and i suppose there will be a PR for doing that merge.

I'm going to base the changes I need on this PR, but note that this will possibility span multiple conflicts if changes are made on this PR, the sooner we can review the better.

jotak:
TBH, I think reviewing the actual code would be easier than reviewing the git diffs from github. Lots of file renaming makes github lost in its diffs.

lib/hawkular/operations/operations_api.rb
jotak: What do you have in memory on provider side, when this function is called?
If you have the whole Entity::Resource object that corresponds to the deployment, we can manage to add its parent_id directly in the model.
If you only have its ID, then yes we can add a REST endpoint to fetch parent.

josejulio: I only have its ID, either adding the parent_id  (so i can fetch the resource) or parent_of (resource_id) works for me. Thanks!

jotak: Done (both options: a rest endpoint + parent_id in model)

jotak:
Note, some tests are currently failing due to missing status information in the latest hawkular-services docker build

jotak:
Is there a reason for not merging?

josejulio:
@israel-hdez can you please review?

lib/hawkular/inventory/entities.rb
```diff
@@ -1,308 +1,121 @@
...
 
-    # Set resource type to the current path
-    # @return a new CanonicalPath based on the current one
-    def resource_type(resource_type)
-      hash = copy_hash
-      hash[:resource_type_id] = hawk_escape_id(resource_type)
-      CanonicalPath.new(hash)
+    def children(recursive = false)
+      return @children unless recursive == true
+      fail Hawkular::ArgumentError 'Resource tree not loaded, load it by calling resource_tree' if @children.nil?
```
israel-hdez, line 101: I don't see this as an argument error.

israel-hdez (approved):
Lots of changes... But it looks OK.

lib/hawkular/inventory/entities.rb
josejulio: I'm open to suggestion/ideas :)

Is something like an `Invalid Call` `Call Error`, `ResourceTreeNotLoadedException` ?

jotak: For information, this is not something that was added in this PR. This PR focuses on removing old inventory, renaming new inv from "v4" to just "inventory" and starting fixing tests, but github diff shows differences between old and renamed files. I think it causes a lot of misunderstanding about this PR. I would have liked it was merged very fast so that we can go on from this basis.

lucasponce:
Confirmed with @jotak this is ready to merge.

PR #240, opened 2017-10-24
[hawkular-1259] Websocket: use dedicated 'senderRequestId' to match req & res

jotak:
Fixes test on concurrent operation 'restart'.

Before, matching WS response with request was done through matching resource ID and operation name. That worked quite fortunately as long as there's no several messages sent concurrently on same resource with same operation name.

Now, it uses a special field that already exists in the CMDGW api, exactly for this purpose: `senderRequestId`.
I think it's much safer, but on another hand, WS cassettes now need to generate a fixed ID in `senderRequestId` so that recorded message and replayed message can match.


jotak:
@israel-hdez , I still have a failing test on JDR export which seems related to the issue you addressed previously: https://github.com/hawkular/hawkular-client-ruby/commit/bb914bf807000cca7329280f5c8168fcf1e725b0

I have two errors there:
1. `expected "..." to respond to blank?`
We talked about it previously (also with @cfcosta ), I don't remember which file I would have to `require` and anyway I don't see why I would need to `require` an extra file if the test worked before with you.

2. `Encoding::UndefinedConversionError:
            "\xAE" from ASCII-8BIT to UTF-8` in `simple_websocket_vcr`, while writing down the record to file. Didn't you had the same issue while recording?

Can you give a little help here please?

israel-hdez:
@jotak To have blank defined, add this line:
`require 'active_support/core_ext/object/blank'`
to `lib/hawkularclient.rb`. That should solve it. Not sure why it's not being required as before.

The error about the encoding... Yes, I had that error, but I solved it in this patch: https://github.com/Jiri-Kremser/simple-websocket-vcr/commit/1d8d6adb52248b8fbc08da7b87be19d11bc6e3f1. Make sure that you have version 0.1.0 of simple-websocket-vcr gem. If you have it, maybe there is another bug on it.

jotak:
@israel-hdez ok, my dependencies were out of date, that was causing the issues. Should be better now.

jotak:
Could it be merged?

israel-hdez:
Any idea about why travis is failing when using the cassettes?
I wouldn't like to merge to master with all the tests failing with vcr.

PR #241, opened 2017-10-24
[hawkular-1259] Using lazy enumerator to fetch inventory

jotak:
@josejulio , @cfcosta can you review please?

One thing that must bring attention is that it is a lazy enumeration, so if the enumeration is looped-over several times, it will trigger hawkular API calls again. So if the consumer plans to iterate several times, it may want to copy it first and iterate over the copy. WDYT?

### DIFF
--- a/lib/hawkular/inventory/inventory_api.rb
+++ b/lib/hawkular/inventory/inventory_api.rb
@@ -13,12 +13,13 @@ module Hawkular::Inventory
     #   http://localhost:8080/hawkular/inventory
     # @param credentials [Hash{String=>String}] Hash of username, password, token(optional)
     # @param options [Hash{String=>String}] Additional rest client options
-    def initialize(entrypoint = nil, credentials = {}, options = {})
+    def initialize(entrypoint = nil, credentials = {}, options = {}, page_size = nil)
       entrypoint = normalize_entrypoint_url entrypoint, 'hawkular/inventory'
       @entrypoint = entrypoint
       super(entrypoint, credentials, options)
       version = fetch_version_and_status['Implementation-Version']
       @version = version.scan(/\d+/).map(&:to_i)
+      @page_size = page_size || 100
     end
 
     # Creates a new Inventory Client
@@ -30,7 +31,7 @@ module Hawkular::Inventory
 
       hash[:credentials] ||= {}
       hash[:options] ||= {}
-      Client.new(hash[:entrypoint], hash[:credentials], hash[:options])
+      Client.new(hash[:entrypoint], hash[:credentials], hash[:options], hash[:page_size])
     end
 
     # Get single resource by id
@@ -61,9 +62,8 @@ module Hawkular::Inventory
     end
 
     # List root resources
-    # @return [Array<Resource>] List of resources
+    # @return [Enumeration<Resource>] Lazy-loaded Enumeration of resources
     def root_resources
-      # FIXME: pagination => lazy-loaded list with ruby?
       resources root: 'true'
     end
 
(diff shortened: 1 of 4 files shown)


lib/hawkular/inventory/inventory_api.rb
```diff
@@ -93,4 +98,24 @@ def fetch_version_and_status
...
+
+  # Lazy fetching results, based on Inventory "ResultSet" model
+  class ResultFetcher
+    include Enumerable
+
+    def initialize(fetcher)
+      @fetcher = fetcher
+    end
+
+    def each
```
josejulio, line 110: Implementing `each` will make sure that `map` and `each` are correctly handled?

cfcosta: As long as you're also including `Enumerable`, you get all those methods for free.

spec/unit/inventory_spec.rb
```diff
@@ -51,4 +51,52 @@
       expect(resource.children_by_type('type-03', true).size).to eq(1)
     end
   end
+
+  describe '#ResultFetcher' do
+    it 'should fetch two first pages' do
+      batches = [[1, 2], [3, 4], [5]]
```
cfcosta: Those variables should be extracted to `let` statements.

spec/unit/inventory_spec.rb
```diff
@@ -51,4 +51,52 @@
       expect(resource.children_by_type('type-03', true).size).to eq(1)
     end
   end
+
+  describe '#ResultFetcher' do
```
cfcosta: If this is a subclass of the inventory API, you can just do `describe ResultFetcher`. `#something` is used to denote instance methods.

spec/unit/inventory_spec.rb
```diff
@@ -51,4 +51,52 @@
...
+    it 'should fetch two first pages' do
+      batches = [[1, 2], [3, 4], [5]]
+      page_size = 2
+      calls_count = 0
+      fetch_func = lambda do |offset|
+        calls_count += 1
+        { 'startOffset' => offset, 'resultSize' => 5, 'results' => batches[offset / page_size] }
+      end
+      result_fetcher = Hawkular::Inventory::ResultFetcher.new(fetch_func)
```
cfcosta: And this to a `subject` block.

spec/unit/inventory_spec.rb
```diff
@@ -51,4 +51,52 @@
       expect(resource.children_by_type('type-03', true).size).to eq(1)
     end
   end
+
+  describe '#ResultFetcher' do
+    it 'should fetch two first pages' do
```
cfcosta: For consistency, those tests should be in the present tense of the third person, as pointed here: http://www.betterspecs.org/#should

```ruby
it 'fetches first two pages'
```

cfcosta (changes requested):

spec/unit/inventory_spec.rb
jotak: Can you detail what you expect?
If I understand let [1], it's meant to be defined on outer-scope for lazy evaluation. In this case lazy evaluation wouldn't bring anything as they are constants (except `calls_count`).
However I can move them to outer scope for less duplicates.

For `calls_count`, I tried to use let: (`let(:calls_count) { 0 }`), but it doesn't like the increment operation: `NoMethodError:
       undefined method '+' for nil:NilClass`.
Btw I think it's good to set it in test scope instead of outer scope, so that the test is self-explanatory.

[1] https://relishapp.com/rspec/rspec-core/v/3-4/docs/helper-methods/let-and-let

spec/unit/inventory_spec.rb
jotak: Can you detail?
From the docs here [1], subject is either used with `its` as a pojo (poro?) to evaluate its fields, or with method `Subject.should`. I can't see how it can be used here, when I need to call `Enumeration.take` with various arguments, or `Enumeration.collect`. How do you suggest I use it?

[1] https://relishapp.com/rspec/rspec-core/v/2-0/docs/subject

jotak:
Could it be merged?

PR #242, opened 2017-10-25
[hawkular-1259] Fixing all tests

jotak:
- Update cassettes with new docker image
- The new docker image fixes inventory/status endpoint issue and brings back authentication by default
- Now using a fixed admin token for tests


jotak:
Note: this PR is based on #240 , which should be merged first, and then this one rebased
Commits displayed here include commits of #240.

jotak:
PS: travis is lying. It works :)

spec/integration/metric_spec.rb
```diff
@@ -661,7 +661,7 @@
             end
 
             data = @client.gauges.get_data(hawkular_mem_id)
-            expect(data.size).to be > 2 # needs the services to be running for more than 2 minutes
+            expect(data.size).to be > 2 # needs the services to be running for some time (~15 min)
```
josejulio: I'm not sure I understand this comment:
`# needs the services to be running for some time (~15 min)`
Yet the live tests are passing in under 10 minutes.

josejulio:
> PS: travis is lying. It works :)

A good kick made it reconsider.

spec/integration/metric_spec.rb
jotak: Interesting... I think collection interval for OS metrics is set to 5 minutes... maybe there's some specific config for travis?

jotak: Just confirming, it's the purpose of script file in `.travis/build_config_for_testing.rb` So, when cassettes are generated locally without any specific configuration, it's about 15 minutes to wait for >2 datapoints of this metric.

jshaughn:
This looks good to me.  Not sure why it is not yet merged as we need to get the 1259 work completed.  I'll merge now, follow-on work can be done in other PR's.

PR #243, opened 2017-10-30
Adds Resource method to search metrics by type

josejulio:
Fetching a hawkular metric type id of a metric

### DIFF
--- a/spec/unit/inventory_spec.rb
+++ b/spec/unit/inventory_spec.rb
@@ -6,16 +6,28 @@ describe 'Inventory' do
     {
       'id' => 'root',
       'type' => {},
+      'metrics' => [
+        { 'type' => 'm1' },
+        { 'type' => 'm2' },
+        { 'type' => 'm3' }
+      ],
       'children' => [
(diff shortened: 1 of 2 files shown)


spec/unit/inventory_spec.rb
```diff
@@ -52,6 +66,31 @@
...
+
+    it 'top level metrics are of type Metric' do
+      expect(resource.metrics).to all(be_a(::Hawkular::Inventory::Metric))
+    end
+
+    it 'all metrics on the tree are of type Metric' do
+      expect(resource.metrics(true)).to all(be_a(::Hawkular::Inventory::Metric))
+    end
+
+    describe '#metrics_by_type' do
```
jotak: Is it intentional to put it inside `describe '#metrics'` block, rather than on the same level?

josejulio: No. My bad.

jotak:
LGTM

josejulio:
Thanks @israel-hdez

PR #244, opened 2017-11-03
Hawkular 1259

jshaughn:
@jotak , Please review.  This is basically the 1259 branch rebased on master.  If you are OK with it I think we can merge (no squash to maintain the commit history).  After the merge we need to update the version to 5.0 in a follow-on PR.


docker-compose.yml
```diff
@@ -1,5 +1,7 @@
 hawkular:
-  image: "hawkular/hawkular-services:latest"
+  # !! WARNING !! FIXME !! TODO !!
```
jotak: So, what's the rightest docker tag now? Is it "hawkular-services:latest", for master?

jotak:
Apart for the docker image to use (in theory it should be "latest", but I have a doubt given its size), it's LGTM in terms of code review. But I let @cfcosta or @israel-hdez tell if it's ok to merge in master.

docker-compose.yml
```diff
@@ -1,5 +1,7 @@
...
+  # Reset to hawkular/hawkular-services:latest before merging in master
+  image: "hawkular/hawkular-services:hawkular-1259"
```
josejulio: I think this can be updated to hawkular-services:latest before we forget about this

jshaughn: pushed fix.

josejulio:
LGTM

jotak:
Thanks @jshaughn

PR #245, opened 2017-11-03
Update to new major version 5.0 for GA work

jshaughn:
@josejulio or @cfcosta Now that we have merged 1259, please review

### DIFF
--- a/CHANGES.rdoc
+++ b/CHANGES.rdoc
@@ -3,6 +3,9 @@
 This document describes the relevant changes between releases of the
 _hawkular-client_ project.
 
+=== v 5.0.0
+* This version introduces support for the new Hawkular Inventory API and Prometheus metric storage.
+
 === v 4.1.0
 * Fixed binary data handling when invoking <em>Export JDR</em> on operations API.
 * Added <em>delete_immediately</em> parameter to <em>Export JDR</em> operation.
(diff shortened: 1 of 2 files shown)


CHANGES.rdoc
```diff
@@ -3,6 +3,9 @@
 This document describes the relevant changes between releases of the
 _hawkular-client_ project.
 
+=== v 5.0.0
+* This version introduces support for the new Hawkular Inventory API and Prometheus metric storage.
```
cfcosta, line 7: This should describe the API changes, is it documented anywhere?

josejulio: Is `Inventory API is completely rewritten, take a look at the new API.`  a valid API change? (or something along those lines)
It works completely different now.

jshaughn: Right, we're not going to document API changes, this is not a progression but a replacement. 
 We aren't really on the hook for backward compatibility between TP3 and GA versions with respect to inventory and metrics.  We could feasibly create a whole new gem although I don't know if that's really necessary as long as we have a solution that works for CM-OPS and their need for an older version.

@cfcosta I only added a section here to get something started, we can revise it as you see fit as we progress.  If there is text you prefer for this PR just enter it into a response and I'll use it verbatim.

israel-hdez: @jshaughn, @josejulio: I see @cfcosta comment as a desire for consistency on our part. The previous inventory iteration was also a rewrite and changes were, at least, minimally mentioned  (see #211)

Besides that... Will 5.0.0 include Prometheus support?

jshaughn: Guys, I am going to close this PR and defer to you for the gem management.

PR #246, opened 2017-11-06
HAWKULAR-1324 inventory model designed for prometheus

jotak:
- Update Inventory.Metric model
- Use new docker image, without hakwular-metrics and cassandra
- Disable hawkular-metrics dependant tests. All disabled tests have been annotated "FIXME"


jotak:
Note: the FIXMEs in code should be fixed via another PR (with JIRA https://issues.jboss.org/browse/HAWKULAR-1293 )

.travis/start_hawkular_services.sh
```diff
@@ -3,21 +3,6 @@
...
-while [ "$CASSANDRA_STATUS" != "running" ] && [ $TOTAL_WAIT -lt 60 ]; do
- CASSANDRA_STATUS=`docker-compose exec myCassandra nodetool statusbinary | tr -dc '[[:print:]]'`
- echo "Cassandra server status: $CASSANDRA_STATUS."
-
- sleep 3
-
- TOTAL_WAIT=$((TOTAL_WAIT+3))
- echo "Waited $TOTAL_WAIT seconds for Cassandra to start."
-done
-
```
josejulio: Good bye Cassandra!

lib/hawkular/metrics/metrics_client.rb
```diff
@@ -59,7 +59,8 @@ def initialize(entrypoint,
                    options = {})
       entrypoint = normalize_entrypoint_url entrypoint, 'hawkular/metrics'
       super(entrypoint, credentials, options)
-      check_version
+      # FIXME: restore for prometheus client
+      # check_version
```
josejulio: We are going to have a separate client for Prometheus, we should leave the hawkular-metrics client as is (or rename to hawkular_metrics)

jotak: I see what you mean... I did that for integration tests to pass, since hawkular-metrics is not part of hawkular-services anymore. I guess we'll have to load some hawkular-metrics standalone image (but, does it exist?) from compose file.

josejulio: I'm sure @rubenvp8510 knows where to find one. I recall he did something for the python client.

rubenvp8510: I have the latest version of metrics on my  dockerhub repo: `rubensvp/hawkular-metrics:latest` it's a pure metrics image.  I'm using it for test `hawkular-python-client`

.travis/start_hawkular_services.sh
jotak: Hello back Cassandra!

jotak:
@josejulio , @rubenvp8510 : I've reset this PR to take back Cassandra & Hawkular Metrics, but the big difference is that now it's a separate thing from the "hawkular client", since they don't share connection settings. In other words, the metrics client needs to be created directly and not through hawkular client.

Another consequence is that some integration tests, that were testings metrics provided by agent/hServices, are not valid anymore so I removed them.

.travis/start_hawkular_services.sh
josejulio: Easy goes, easy comes.

lib/hawkular/inventory/entities.rb
```diff
@@ -117,8 +108,8 @@ def metrics(recursive = false)
       children(recursive).collect(&:metrics).flat_map(&:itself).concat(@metrics)
     end
 
-    def metrics_by_type(type)
-      @metrics.select { |m| m.type == type }
+    def metrics_by_family(family)
+      @metrics.select { |m| m.family == family }
```
josejulio, line 112: What exactly a family is? Is akin to a meta type? e.g. servers (widlfly 10, eap 6 , eap 7, jdg) ?

jotak: No, it's the "family name" in prometheus, for instance a metric with ID `jvm_memory_pool_bytes_used{host=locahost,feedid=1234}` has family name `jvm_memory_pool_bytes_used`.
I'm not sure yet if we're going to be able to get availability metrics from family names, I think it's something @lucasponce will face with the task on hawkular-provider.

jmazzitelli: There are nothing called "availability metric" in the new inventory or Prometheus itself. At least, it isn't explicitly called that (i.e. there is not a first-class concept of an availability metric like we have in Hawkular Metrics). You just 'have to know' a specific metric corresponds to UP or DOWN. For example:

https://github.com/hawkular/hawkular-commons/blob/master/hawkular-inventory-parent/hawkular-inventory-service/src/main/resources/wildfly-10-jmx-exporter.yml#L72-L86

```
- pattern: "^jboss.as<deployment=(.+)><>status: OK"
    attrNameSnakeCase: true
    name: wildfly_deployment_availability
    value: 1
    type: GAUGE
    labels:
      name: $1
 - pattern: "^jboss.as<deployment=(.+)><>status: (FAILED|STOPPED)"
    attrNameSnakeCase: true
    name: wildfly_deployment_availability
    value: 0
    type: GAUGE
    labels:
      name: $1
```

This is a "gauge" metric in Prometheus. We just happen to define it to have a value of 1 if the deployment's status is "OK" and 0 if the deployment's status is "FAILED" or "STOPPED". You just have to know the gauge metric that we assign the family name "wildfly_deployment_availability" is really a 0 or 1 "availability" metric.

PR #247, opened 2017-11-14
[Hawkular-1259] Preparation to release version 5.0.0.pre1

israel-hdez:
* Describing high level changes in the version
* Describing breaking changes
* Bumping version in gemspec

5.0.0.pre1 will include [HAWKULAR-1259](https://issues.jboss.org/browse/HAWKULAR-1259) work.

### DIFF
--- a/hawkularclient.gemspec
+++ b/hawkularclient.gemspec
@@ -11,7 +11,7 @@ Gem::Specification.new do |gem|
   gem.homepage      = 'https://github.com/hawkular/hawkular-client-ruby'
   gem.summary       = 'A Ruby client for Hawkular'
   gem.license       = 'Apache-2.0'
-  gem.required_ruby_version = '>= 2.0.0'
+  gem.required_ruby_version = '>= 2.2.0'
   gem.description = <<-EOS
     A Ruby client for Hawkular
   EOS
--- a/CHANGES.rdoc
+++ b/CHANGES.rdoc
@@ -3,6 +3,20 @@
 This document describes the relevant changes between releases of the
 _hawkular-client_ project.
 
+=== v 5.0.0.pre1
+* Hawkular services version 0.40.x and Hawkular Agent 1.x will be the last
+  versions to include support for _inventory-on-metrics_ and Hawkular Metrics.
+  The following versions won't be backwards compatible. Right now, a new API
+  for inventory is already implemented. This pre-release of the client is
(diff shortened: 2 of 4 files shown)


jshaughn:
Great, happy to see the new version getting ready to be released.  LGTM.

hawkularclient.gemspec
```diff
@@ -11,7 +11,7 @@ Gem::Specification.new do |gem|
   gem.homepage      = 'https://github.com/hawkular/hawkular-client-ruby'
   gem.summary       = 'A Ruby client for Hawkular'
   gem.license       = 'Apache-2.0'
-  gem.required_ruby_version = '>= 2.0.0'
+  gem.required_ruby_version = ['>= 2.2.0', '< 2.5']
```
cfcosta: Good for updating from `2.0.0`, but I think we should avoid adding an upper-limit to the ruby version. New ruby versions very rarely break stuff, and this might break setups where the user might want to upgrade ruby without upgrading gem versions.

cfcosta (changes requested):

josejulio (approved):
LGTM

hawkularclient.gemspec
josejulio: Any reason why `2.5` is the upper limit?

lucasponce:
I am working also in
https://issues.jboss.org/browse/HAWKULAR-1294
https://issues.jboss.org/browse/HAWKULAR-1293
and I will need also to add changes into the client.

So, just to confirm, 5.0.0.pre1 is intended to work with the new versions of Hawkular-Services ?
Does consolidates all work done for hawkular-1259 ?

jshaughn:
@israel-hdez I think it would be more clear if this PR were linked to a Jira, perhaps you can edit the title?

josejulio:
> So, just to confirm, 5.0.0.pre1 is intended to work with the new versions of Hawkular-Services ?
> Does consolidates all work done for hawkular-1259 ?

This is intended to work with hawkular-1259

israel-hdez:
@jshaughn @lucasponce  I updated title and description.

hawkularclient.gemspec
israel-hdez: Just because 2.5 still does not exist. But I'll take @cfcosta suggestion and not place the upper bound.

cfcosta (approved):

PR #248, opened 2017-11-15
Fixes on Metrics v8 tests

lucasponce:


jotak:
LGTM, just waiting for travis status.

josejulio:
are the tests supposed to pass at this point? They are failing on the live server.

lucasponce:
> are the tests supposed to pass at this point? They are failing on the live server.

I see latest commits were not passing, and I think docker is not yet in sync-ed.
I guess that can be done in a following step.

PR #249, opened 2017-11-16
[WIP][HAWKULAR-1293] Add a Prometheus client

lucasponce:
Please don't merge yet. This is a WIP and it would need a counterpart in MiQ repos.
I will add additional api methods as long as we need it.


lib/hawkular/prometheus/prometheus_api.rb
```diff
@@ -0,0 +1,61 @@
...
+        metrics.each do |metric|
+          if contains_metric_in_result(metric, result)
+            result['metric'] = metric
+            break
+          end
+        end
+      end
+    end
+
+    def contains_metric_in_result(metric, result)
```
jshaughn: I think the convention for predicate naming would be contains_metric_in_result?

spec/integration/prometheus_spec.rb
```diff
@@ -0,0 +1,46 @@
+require_relative '../vcr/vcr_setup'
+require_relative '../spec_helper'
+
+module Hawkular::Prometheus::RSpec
```
cfcosta, line 4: This is not used on the specs (it is in some places, but should be removed). The correct way to describe tests is:

```ruby
describe Hawkular::Prometheus do
  # tests
end
```

spec/integration/prometheus_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      description = example.description
+      description
+    end
+
+    around(:each) do |example|
+      record('Prometheus/Queries', credentials, cassette_name, example: example)
+    end
+
+    before(:each) do
+      @client = Hawkular::Prometheus::Client.new(HAWKULAR_BASE, creds, options)
```
cfcosta, line 20: Use a `let` definition instead of defining instance variables.

spec/integration/prometheus_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+    it 'Should Fetch Metrics' do
```
cfcosta: Missing a new line here.
Also, tests should be described as phrases in the third person, current tense. So, this is good:

```ruby
it 'fetches metrics' do
```

While those are bad:

```ruby
it 'should fetch metrics'
it 'Should Fetch Metrics'
```

The reason for that is because of the `specification` formatter for RSpec, that generates a textual description of what the test suite is tested, and in proper codebases is also a specification on what the system does.

Check http://www.betterspecs.org/#should for reference.

spec/integration/prometheus_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+                                        'expression' => 'jvm_memory_bytes_used{area="nonheap",feed_id="attilan"}',
+                                        'labels' => { area: 'nonheap', feed_id: 'attilan' })
+      ]
+
+      results = @client.query_range(metrics: metrics,
+                                    starts: '2017-11-21T09:00:00Z',
+                                    ends: '2017-11-21T11:00:00Z',
+                                    step: '5s')
+
+      expect(results.map { |r| r['metric'].name }).to include('Heap Used', 'NonHeap Used')
```
cfcosta: Here I'd test each one of them individually, like so:

```ruby
first = results.first
second = results.last

expect(first['metric'].name).to eq 'Heap Used'
expect(second['metric'].name).to eq 'NonHeap Used'
```

The less processing we do on each test, the better.

lib/hawkular/hawkular_client.rb
```diff
@@ -2,11 +2,12 @@
 require 'hawkular/alerts/alerts_api'
 require 'hawkular/tokens/tokens_api'
 require 'hawkular/operations/operations_api'
+require 'hawkular/prometheus/prometheus_api'
```
cfcosta, line 5: I don't like this to be honest. To me it seems like the metrics endpoint should be pluggable, so exposing the metrics provider underlying is binding us on each codebase to that provider, making changes harder. To me, we should just provide a `metrics` endpoint, as a proxy object for each provider (be it hawkular-metrics, be it prometheus).

lib/hawkular/prometheus/prometheus_api.rb
```diff
@@ -0,0 +1,50 @@
+require 'hawkular/base_client'
+require 'hawkular/inventory/entities'
+require 'ostruct'
+
+module Hawkular::Prometheus
+  class AlerterClient < Hawkular::BaseClient
```
cfcosta: The `Client` part in this class name seems unnecessary, given that `Alerter` is already the role of the class.

lib/hawkular/prometheus/prometheus_api.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+  end
+
+  # Interface to talk with the Prometheus server used for Middleware Manager
+  # @param entrypoint [String] base url of Hawkular Services
+  class Client < Hawkular::BaseClient
+    def initialize(entrypoint, credentials = {}, options = {})
+      prometheus_entrypoint = AlerterClient.new(entrypoint, credentials, options).prometheus_entrypoint
+      entrypoint = normalize_entrypoint_url prometheus_entrypoint, 'api/v1'
+      @entrypoint = entrypoint
```
cfcosta: Use parens on methods that are included in an expression, on a first glance it was really hard to understand this because it reads like one expression instead of a method with two parameters.

cfcosta: `@entrypoint` is set by `BaseClient`, right? I wouldn't also set it here, then.

If not, at least there's no need to set the temporary variable, so:

```ruby
@entrypoint = normalize_entrypoint_url(prometheus_entrypoint, 'api/v1')
```

lib/hawkular/prometheus/prometheus_api.rb
```diff
@@ -0,0 +1,50 @@
...
+    def prometheus_entrypoint
+      rest_client('/prometheus/endpoint').get
+    end
+  end
+
+  # Interface to talk with the Prometheus server used for Middleware Manager
+  # @param entrypoint [String] base url of Hawkular Services
+  class Client < Hawkular::BaseClient
+    def initialize(entrypoint, credentials = {}, options = {})
+      prometheus_entrypoint = AlerterClient.new(entrypoint, credentials, options).prometheus_entrypoint
```
cfcosta: Doesn't seem to me like the `Alerter` should be responsible to define the prometheus entrypoint. It should, imho, be defined on the client itself.

lib/hawkular/prometheus/prometheus_api.rb
```diff
@@ -0,0 +1,50 @@
...
+      query = metrics.map(&:expression).join(' or ')
+      response = http_get "/query_range?start=#{starts}&end=#{ends}&step=#{step}&query=#{query}"
+      results = response['data']['result']
+      attach_metric_in_result(results, metrics)
+    end
+
+    private
+
+    def attach_metric_in_result(results, metrics)
+      results.each_with_index { |result, index| result['metric'] = metrics.fetch(index) }
```
cfcosta: This seems... weird. First weird thing is that we are returning hashes instead of objects, which is different from `Inventory`, for example. Second, it seems extremely contrived to do this, but to be honest I can not think of a better option right now, maybe something with `Enumerable#zip`?

cfcosta (changes requested):

lib/hawkular/hawkular_client.rb
lucasponce: At the moment we need both in parallel as the same gem can be used with hMetrics and P8S

lib/hawkular/prometheus/prometheus_api.rb
lucasponce: To be honest, I don't mind but I would mantain to indicate that this is a class which is client for an endpoint.

lib/hawkular/prometheus/prometheus_api.rb
lucasponce: From users perspective, P8S is given by MW Manager. MW Manager resolves the endpoint where P8S is hosted, so, this is just the flow to resolve this.

lib/hawkular/prometheus/prometheus_api.rb
lucasponce: I guess I am just following the same pattern used in other clients.
Probably your comment is right, but I would try to be consistent in the meaning that potential refactoring in whole client could go in a separate PR.

lucasponce: +1

spec/integration/prometheus_spec.rb
lucasponce: If all spec files are using this pattern, then this is a refactoring.
Which is good, but I think it should go in a different PR not on this.

lib/hawkular/hawkular_client.rb
cfcosta: Exactly the reason why I think it should be the same endpoint with different providers.

lib/hawkular/prometheus/prometheus_api.rb
cfcosta: Disagree, I think we should get this one correct and refactor the other clients on a separate PR.

lib/hawkular/prometheus/prometheus_api.rb
lucasponce: The feature of this is to attach the metric model into the results which is what is needed for the provider's logic.

spec/integration/prometheus_spec.rb
lucasponce: +1

spec/integration/prometheus_spec.rb
cfcosta: Same as the other comment, this should be correct, given that it is new code. Another PR can be made to refactor the other ones.

spec/integration/prometheus_spec.rb
lucasponce: This is good, but it is the same, I will use the same pattern used in the other specs files and we can refactor later in a separate PR.

spec/integration/prometheus_spec.rb
lucasponce: +1 
I like this approach, I will change it, thx.

lucasponce:
@cfcosta thx for your comments.
Two things, this is a WIP, and it is not yet ready for review.
I will try to add your comments, but I would not include refactoring of style for consistency with other files in the client.
Those refactors are good but I think they should go in a separate PR to not mix with this logic.

lib/hawkular/hawkular_client.rb
lucasponce: No, as they have different models behind.
There is not a pure abstraction model plugable where provider can be hidden.

lucasponce:
I'm going to close this PR, as I am going to consolidate changes from a different branch.

PR #251, opened 2017-11-30
Fix to_camel_case

jshaughn:
Fix to_camel_case such that it does not downcase a param that is already camel case.

### DIFF
--- a/lib/hawkular/operations/operations_api.rb
+++ b/lib/hawkular/operations/operations_api.rb
@@ -431,8 +431,14 @@ module Hawkular::Operations
     end
 
     def to_camel_case(str)
-      ret = str.split('_').collect(&:capitalize).join
-      ret[0, 1].downcase + ret[1..-1]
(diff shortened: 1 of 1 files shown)


lib/hawkular/operations/operations_api.rb
```diff
@@ -431,8 +431,14 @@ def prepare_payload_hash(ignored_params, hash)
     end
 
     def to_camel_case(str)
-      ret = str.split('_').collect(&:capitalize).join
-      ret[0, 1].downcase + ret[1..-1]
+      subs = str.split('_')
+      if subs.length > 1
+        ret = subs.collect(&:capitalize).join if subs.length > 1
```
josejulio: Isn't this second `if subs.length > 1` redundant ?

jshaughn: whoops, my bad, good eye.

josejulio:
LGTM

PR #254, opened 2017-12-05
Addition on Prometheus API and fix specs

lucasponce:
Added a trivial ping method to be used from the provider.
Also, modified the Specs to simplify the use on cassettes and live servers.
This is tricky as on live servers we depend of time range, to not add complexity, we will check basic uses.

@israel-hdez the following changes to https://github.com/israel-hdez/manageiq-providers-hawkular/pull/5 will need this one.

One thing missing is that for live tests it needs to setup the proper hServices setup with the Prometheus server.

I guess we will need something similar like
https://github.com/hawkular/hawkular-services/tree/hawkular-1275/dist/src/test/scripts

I haven't yet started that, I am going to prioritize the changes and the specs on cassettes.

Probably to adjust the live tests will be pending before I go to PTO tomorrow, so feel free is you can work on that.

/cc @jshaughn


lucasponce:
Ok, local cassettes passes but there is a pending change on the live tests

jshaughn:
LGTM

PR #255, opened 2017-12-12
Return nil in "resource" method if requested resource doesn't exist

israel-hdez:
Method `resource` of inventory API is returning the generic
`Hawkular::Exception` if any error occurs. However, the special case of
querying non existent resources should be easily identifiable.

To accomplish this, a a nil value is returned if Hawkular replies with
404 status.


josejulio:
LGTM

